PR #8, opened 2023-05-05
Make: Windows Compatibility

ishkhan42:


include/usearch/usearch.hpp
```diff
@@ -479,8 +512,11 @@ class usearch_pack_m uint40_t {
         return old;
     }
 };
+#ifdef WINDOWS
+__pragma(pack(push, 1))
+#endif // WINDOWS
 
-static_assert(sizeof(uint40_t) == 5, "uint40_t must be exactly 5 bytes");
+    static_assert(sizeof(uint40_t) == 5, "uint40_t must be exactly 5 bytes");
```
ashvardanian, line 519: What's up with the indent?

include/usearch/usearch.hpp
```diff
@@ -412,7 +422,12 @@ class max_heap_gt {
  *
  */
 class mutex_t {
+#ifdef WINDOWS
```
ashvardanian, line 425: `#if defined(WINDOWS)`

src/bench.cpp
```diff
@@ -3,12 +3,26 @@
...
+#include <Windows.h>
+
+#pragma comment(lib, "Dbghelp.lib")
+
+#define STDERR_FILENO HANDLE(2)
+#else
 #include <execinfo.h>
-#include <fcntl.h>    // `open`
-#include <stdlib.h>   // `getenv`
+#include <fcntl.h>    // `fallocate`
```
ashvardanian, line 19: We don't use `fallocate`, revert previous annotation.

include/usearch/usearch.hpp
```diff
@@ -7,26 +7,39 @@
...
-#include <cmath>     // `std::sqrt`
-#include <cstring>   // `std::memset`
-#include <mutex>     // `std::unique_lock` - replacement candidate
-#include <random>    // `std::default_random_engine` - replacement candidate
-#include <unistd.h>  // `open`, `close`
-#include <utility>   // `std::exchange`
-#include <vector>    // `std::vector`
 
 #if defined(WIN32) || defined(_WIN32) || defined(__WIN32__) || defined(__NT__)
+#define NOMINMAX // define this macro to prevent the definition of min/max macros in Windows.h
```
ashvardanian, line 12: This will affect compiling every file where `usearch.hpp` is included. There is a way without defining it. We may want to provide our own `min` and `max` instead.

ashvardanian (changes requested):

ashvardanian:
Aside from compilation fixes, @ishkhan42, we also need to extend CI to produce packages for Windows as well.

include/usearch/usearch.hpp
ishkhan42: We define our own `min` and `max` functions, which is why this macro is required. There is also a conflict regarding that in Clipp.

ashvardanian: Where do we define the `min` and the `max`? The `clipp` isn't a problem, cause we only include it in the benchmark executable, while this is a library header, that others will pull.

include/usearch/usearch.hpp
ishkhan42: Why?

include/usearch/usearch.hpp
ashvardanian: Oh, I see. That function is never used, and I removed it. Please fetch the newest version from `main-dev`.

include/usearch/usearch.hpp
ashvardanian: Because that's our style everywhere else. It makes searching easier.

include/usearch/usearch.hpp
```diff
@@ -7,26 +7,39 @@
...
+
+#define usearch_pack_m
+#define usearch_align_m __declspec(align(64))
+#define WINDOWS
+
 #else
 #include <fcntl.h>    // `fallocate`
 #include <stdlib.h>   // `posix_memalign`
 #include <sys/mman.h> // `mmap`
-#include <sys/stat.h> // `fstat` for file size
```
ashvardanian, line 27: Don't we need `<sys/stat.h>` for `fstat`?

ishkhan42: It's not removed

ashvardanian (approved):

PR #11, opened 2023-05-09
Add: GoLang bindings

ashvardanian:


golang/golang.cpp
```diff
@@ -0,0 +1,124 @@
+#include "../src/advanced.hpp"
+#include <algorithm>
```
ashvardanian, line 2: This is probably not needed.

golang/golang.cpp
```diff
@@ -0,0 +1,124 @@
...
+    // translate into sefaults and are not caught by the runtime
+    try {
+        // todo:: the code cannot currently be used in a multithreaded environment
+        // because of golang's roaming gorountines, but once we have a way to
+        // work around it, we should revisit this as this check has a toctou race
+        if (index->size() + 1 >= index->capacity())
+            index->reserve(ceil2(index->size() + 1));
+        index->add(label, vector);
+    } catch (std::exception& e) {
+        return e.what();
```
ashvardanian, line 103: This isn't safe, as you are returning a pointer to the string that may already be destroyed.

golang/golang.cpp
```diff
@@ -0,0 +1,124 @@
...
+    }
+    return NULL;
+}
+
+SearchResults search(native_index_t* index, float* query, int query_len, int limit) {
+    // todo:: this could be allocated as golang slice
+    // to avoid a copy. not sure how it interacts with gc
+    // that is why doing this now
+    label_t* matches_data = new label_t[limit];
```
ashvardanian, line 112: Yes, a `Slice` is better.

golang/lib.go
```diff
@@ -0,0 +1,306 @@
...
+	"sync"
+	"unsafe"
+)
+
+/*
+#include <stdlib.h>
+#include "golang.h"
+// avoid some verbosity by using void* index here
+// proper typed pointer in the cpp file definitions
+void* new_index(char* metric_str, int metric_len,
```
ashvardanian, line 16: I would generally recommend adding a prefix to function names, like `usearch_` to avoid collisions with more common verbs, aka `save`, `size`, `capacity`...

golang/lib.go
```diff
@@ -0,0 +1,306 @@
...
+type Index struct {
+	opeque_handle *C.void
+	config        IndexConfig
+}
+
+func NewIndex(conf IndexConfig) *Index {
+	ind := &Index{config: conf}
+	// make sure that all Index operations go through the same OS thread
+	// as the C code uses thread local storage
+	runtime.LockOSThread()
```
ashvardanian, line 129: Where are we using a `thread_local` storage?

golang/lib.go
```diff
@@ -0,0 +1,306 @@
...
+
+func (ind *Index) VecDimension() int {
+	return int(C.dimensions(unsafe.Pointer(ind.opeque_handle)))
+}
+
+func (ind *Index) Capacity() int {
+	return int(C.capacity(unsafe.Pointer(ind.opeque_handle)))
+}
+
+func (ind *Index) SetCapacity(capacity int) error {
```
ashvardanian, line 216: What's the convention? In other languages this is generally called `reserve`.

golang/lib.go
```diff
@@ -0,0 +1,306 @@
...
+	case L2_SQ:
+		return "l2_sq"
+	case IP:
+		return "ip"
+	case COS:
+		return "cos"
+	case Haversine:
+		return "haversine"
+	default:
+		panic("unknown metric")
```
ashvardanian, line 65: We can start with `panic`, but returning `error` is better. Check out the [UStore bindings](https://github.com/unum-cloud/go-ustore/blob/main/internal/ukv.go) for inspiration.

PR #25, opened 2023-05-16
Make: Use .releaserc for semantic-release

ishkhan42:


.github/workflows/update_version.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+echo $1 > VERSION && 
+    sed -i "s/\"version\": \".*\"/\"version\": \"$1\"/" .vscode/launch.json &&
```
ashvardanian, line 4: This is the version of VSCode file, not USearch.

ashvardanian (changes requested):

PR #37, opened 2023-05-19
Server client fixes 

ishkhan42:
Minor fixes in client.py and server.py.
Minor Refactor in release.yml.


ashvardanian:
Thank you! This is precisely the kind of bug we would prevent with [this](https://github.com/unum-cloud/ucall/issues/64#issuecomment-1542876172).

ashvardanian (approved):

PR #51, opened 2023-05-23
Fix: Taking address of packed member  is UB

ishkhan42:
[Taking address of packed member of 'Packed' may result in an unaligned pointer value](https://stackoverflow.com/a/59764793)


ashvardanian:
This isn’t a fix, it just disables the alignment.

PR #111, opened 2023-06-16
Refactor: replace dataclasses with namedtuples

Arman-Ghazaryan:


ashvardanian:
`dataclasses` are only present in Python 3.7. Your commit doesn't replace them entirely: https://github.com/unum-cloud/usearch/blob/8d86efdfe124a03deb33c47fd85dabc12a994468/python/usearch/eval.py#L2

PR #129, opened 2023-06-26
Add: WASM bindings

DarvinHarutyunyan:
Add: "USearch WASM Build" GitHub action


cpp/bench.cpp
```diff
@@ -625,12 +637,14 @@ int main(int argc, char** argv) {
         exit(0);
     }
 
-    // Instead of relying on `multithreaded` from "index_punned_dense.hpp" we will use OpenMP
-    // to better estimate statistics between tasks batches, without having to recreate
-    // the threads.
+// Instead of relying on `multithreaded` from "index_punned_dense.hpp" we will use OpenMP
+// to better estimate statistics between tasks batches, without having to recreate
+// the threads.
+#if USEARCH_USE_OPENMP
```
ashvardanian, line 643: Putting macros under comments breaks formatting.

ashvardanian (changes requested):

ashvardanian:
The issue is in the USearch design targeting 64-bit architectures vs. Emscripten and WebAssembly being 32-bit by default. This can be seen even [by `v |= v >> 32` warnings](https://github.com/unum-cloud/usearch/actions/runs/5388497657/jobs/9781237440?pr=129#step:5:49). There are a few ongoing discussions around a 64-bit runtime: [1](https://github.com/emscripten-core/emscripten/issues/12087), [2](https://github.com/WebAssembly/memory64). A good plan may be adding a 32-bit test and gradually fixing the upcoming issues, further increasing applicability.

ashvardanian:
I also suggest rebasing from `main-dev` to get some patches.

ashvardanian:
Using JeMalloc and requiring G++-12 for WASM builds is a bad idea. Similarly, SimSIMD doesn't provide any overloads for WebAssembly. I'd suggest removing it from the current build to simplify the process and extend SimSIMD to support WASM intrinsics. 

The dot-product and cosine distance for f32 vectors would be pretty easy to implement [with `wasm_f32x4_add` and `wasm_f32x4_mul`](https://emscripten.org/docs/porting/simd.html).

ashvardanian:
@DarvinHarutyunyan, can you please add an example of using the WASM build?
@gurgenyegoryan I have preset the environment variable `WASMER_TOKEN` to publish the package to `wasmer.io`. Please update our `release.yml` to publish the binary and to update the `wasmer.toml` version number.

ashvardanian (approved):

PR #139, opened 2023-07-10
Add: Usearch build and test GitHub action

gurgenyegoryan:


ashvardanian (changes requested):
Can you rename commits from `Add` to `Make`? Otherwise, it triggers a minor release just for those CI actions. Can you also do rigorous ASAN checks?

ashvardanian:
@gurgenyegoryan I believe ASAN should also be enabled for GCC builds, as a common alternative.

gurgenyegoryan:
> @gurgenyegoryan I believe ASAN should also be enabled for GCC builds, as a common alternative.

@ashvardanian Yes, it's enabled for GCC too.

Issue #161, opened 2023-07-24
Feature: Parallel `view`


ashvardanian:
### Describe what you are looking for

Working with large datasets, it's a pity when your 200+ vCPU machine is idle, waiting for one thread to parse all the files.
Let's adjust `view` function to use the `executor` passed. It will require structural changes to the file, to be able to jump from one offset to another.

### Can you contribute to the implementation?

- [x] I can contribute

### Is your feature request specific to a certain interface?

It applies to everything

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #159, opened 2023-07-21
Invalid: Memory alignment


ashvardanian:
### Describe the bug

There is a data alignment issue, that we have accounted for ahead of time, but haven't implemented the solution yet.
The `index_config_t::vector_alignment` controls, how to store vectors of `add`-ed elements in allocated memory.
Some platforms won't have an issue loading an 8-byte `double`, with an address that is not a multiple of 8.
Others will have that issue.
So the `index_gt` class has to be patched.

### Steps to reproduce

Run the C++ unit test (available in VS code launchers) with `-fsanitize=alignment` address sanitizer flag set.

### Expected behavior

No errors or warnings.

### USearch version

v0.22.3

### Operating System

Any

### Hardware architecture

x86

### Which interface are you using?

C++ implementation

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ashvardanian:
This was resolved by implementing a completely new serialization form and using separate memory allocators for vectors and neighbors lists.

PR #162, opened 2023-07-24
Add: Vector Alignment

mgevor:


ashvardanian:
It's best not to store any additional info in the `node_bytes_split_t`. It should resemble the `node_t`. There should be an easy way to recompute those on-the-fly. We can try implementing that for #159 and parallel views of #161 as part of the same PR.

ashvardanian (approved):

PR #165, opened 2023-07-26
Fix: Typo in metric name

AleksandrKent:


python/usearch/index.py
```diff
@@ -199,7 +199,7 @@ def __init__(
         :param metric: Distance function, defaults to MetricKind.IP
         :type metric: Union[MetricKind, Callable, str], optional
             Kind of the distance function, or the Numba `cfunc` JIT-compiled object.
-            Possible `MetricKind` values: IP, Cos, L2sq, Haversine, Pearson,
+            Possible `MetricKind` values: IP, Cos, L2_sq, Haversine, Pearson,
```
AleksandrKent, line 202: Maybe I mad a mistake to correct MetricKind values of L2sq to L2_sq at this line. I thought this should match `str` values for metrics.

ashvardanian: It’s either PascalCase or snake_case. Enums are generally the first, strings are the second in most Python repos.

AleksandrKent: @ashvardanian okay, there is still other typo in `_normalize_metric` in my commit above.

ashvardanian (approved):

PR #170, opened 2023-07-31
Add compile time feature-gate for lookups and removals

Ngalstyan4:
I am not sure you guys want this in upstream usearch. But, we use this for the following:

For our use-case, we only need vector insertions and search.
Currently, usearch **always** builds a lookup table for vector labels at index creation time (e.g. [here](https://github.com/unum-cloud/usearch/blob/main/include/usearch/index_punned_dense.hpp#L405-L406)) to support `contains, get, remove` queries. I think there are many use-cases that do not need these (including our use-case) so it would be good to not pay the cpu and memory cost of this lookup table creation.

This PR feature-gates the index creation functions and all the methods depending on them so one can easily disable them and remove their runtime cost.

Perhaps the one-time lookup table creation cost and the small maintenance cost is trivial for most use-cases. But, for our case, we need to re-initialize the usearch context for each search query so it is crucial to minimize its memory footprint. This is likely not the typical case for usearch, but this is one case to consider.

I considered leaving the code in and just adding a runtime flag to enable lookups. But preferred this approach because with this the compiler and the IDE did more to help ensure available APIs are used.

### DIFF
--- a/include/usearch/index_punned_dense.hpp
+++ b/include/usearch/index_punned_dense.hpp
@@ -185,12 +185,14 @@ class index_punned_dense_gt {
     using shared_lock_t = std::unique_lock<shared_mutex_t>;
     using unique_lock_t = std::unique_lock<shared_mutex_t>;
 
+#if USEARCH_LOOKUP_LABEL
     mutable shared_mutex_t labeled_lookup_mutex_;
     tsl::robin_map<label_t, id_t> labeled_lookup_;
 
     mutable std::mutex free_ids_mutex_;
     ring_gt<id_t> free_ids_;
     label_t free_label_;
+#endif
 
   public:
     using search_result_t = typename index_t::search_result_t;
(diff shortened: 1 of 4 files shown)


include/usearch/index_punned_dense.hpp
```diff
@@ -609,12 +638,14 @@ class index_punned_dense_gt {
         auto typed_result = typed_->copy(config);
         if (!typed_result)
             return result.failed(std::move(typed_result.error));
+#if USEARCH_LOOKUP_LABEL
         if (!result.index.free_ids_.reserve(free_ids_.size()))
             return result.failed(std::move(typed_result.error));
         for (std::size_t i = 0; i != free_ids_.size(); ++i)
             result.index.free_ids_.push(free_ids_[i]);
 
         result.index.labeled_lookup_ = labeled_lookup_;
```
Ngalstyan4, line 647: where would copy be used? 
Can the old index be used while/after copying? 
if so, should this not keep track of the old lookup table lock as well?

ashvardanian:
Thanks, @Ngalstyan4! It’s a valid suggestion. We can also solve it differently, by adding a single template argument. Will discuss that today ;)

ashvardanian:
@Ngalstyan4, just to clarify, memory savings from removing the lookup table are more important than the ability to remove entries?

Ngalstyan4:
Right

PR #171, opened 2023-07-31
Add external retriever to usearch so vector nodes can be externally stored and managed

Ngalstyan4:
This is incomplete but hopefully has enough to clarify the approach.
What's missing?
- [ ] Allow setting external retriever functions in `usearch_init`, via the configuration struct
- [ ] allow passing an optional opaque pointer in `set_retriever` that will be passed to the external retrievers
- [ ] Wherever there are sets of functions with different type parameters, currently only the functions, currently only the functions corresponding to f32 are properly instrumented.

Note: this is based off of the [lookup label PR](https://github.com/unum-cloud/usearch/pull/170) so changes from that PR also appear here.

### DIFF
--- a/c/lib.cpp
+++ b/c/lib.cpp
@@ -59,6 +59,7 @@ add_result_t add_(index_t* index, usearch_label_t label, void const* vector, sca
     }
 }
 
+#if USEARCH_LOOKUP_LABEL
 bool get_(index_t* index, label_t label, void* vector, scalar_kind_t kind) {
     switch (kind) {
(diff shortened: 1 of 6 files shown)


ashvardanian:
Thanks again, @Ngalstyan4! Our plans seem well aligned. In terms of implementation, there may be a better alternative using batch-evaluated metrics expected in 1.0.0. Will share more details soon.

Ngalstyan4:
more up to date implementation of this is in #335

PR #173, opened 2023-07-31
Add: Batch add

davvard:


ashvardanian:
It's better to use a lambda to invoke the insertion on every input entry rather than pushing them to `std::vector`-s. Let's refactor for efficiency.

davvard:
@ashvardanian Yes you are right, it's already done.

Issue #160, opened 2023-07-21
Feature: C# bindings and Semantic Kernel integration


ashvardanian:
### Describe what you are looking for

With most of the common programming languages covered, the only major one we don't have bindings for is C#.
Having that will open doors to integrate USearch into Semantic Kernel C# backend, as well as a number of Unity applications.

### Can you contribute to the implementation?

- [X] I can contribute

### Is your feature request specific to a certain interface?

Other bindings

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ashvardanian:
A couple of recommendations from @dluc:

> 1. I would encourage adopting the default C# code style (settings can be found in the SK repo, e.g. see https://github.com/microsoft/semantic-kernel/blob/main/.editorconfig)
> 2. if possible please make the library .NET Standard 2.0 compatible; that will increase adoption, e.g. across some old apps that haven't migrated to Standard 2.1 or .NET6+

dluc:
Looking forward to this, in particular the disk based index feature. Let me know if I can help! :-)

ashvardanian:
@AleksandrKent has already made some [great progress on the Python side](https://github.com/microsoft/semantic-kernel/compare/main...AleksandrKent:semantic-kernel:usearch) . C# next. 

@dluc, for compatibility and simplicity, we currently put metadata into Parquet files on disk, but integrating UStore down the road would make more sense.

dranger003:
Hi there, I've taken an initial take on some C# bindings, feel free to use any of it if you think this can be useful.
https://github.com/dranger003/usearch-dotnet

ashvardanian:
Oh, wow! Great to have you with us @dranger003! We are also half way through with those, probably makes sense to syn up and merge the progress. @AleksandrKent how is your fork doing?

ashvardanian:
@dranger003 would you like to merge your progress into the main repo under the `csharp/` directory?

ashvardanian:
Thank you, @dranger003! We will continue from there and will release a NuGet package on GitHub.

dranger003:
Hi @AleksandrKent, great work on interop code! I only focused on the low level P/Invoke calls, which have been pulled. Maybe you can sync up and adjust your wrapper? There are few more things I think I need to clean in the interop layer:
[...]

PR #190, opened 2023-08-06
C# bindings projects re-structure and add auto nuget packages generation for native and interop.

dranger003:
This is in response to https://github.com/unum-cloud/usearch/issues/160#issuecomment-1666718651

The re-structure will automatically produce nuget packages upon compiling (given you have `nuget.exe` in path):
- Packages will be saved under the `nuget` folder
- There are two packages per platform (`x86`/`x64`) one for native and one for interop (four in total for now)
- Current configuration supports only Windows but Linux should be straightforward to add (need to generate two more packages for native as interop already supports Linux)
- You will need to update the `nuspec` files and the interop `csproj` to edit your nuget details (i.e. authors, company, license, etc.)

Also, in `index.hpp` on following line you must cast zero for an x86 build otherwise the overload is ambiguous:
https://github.com/unum-cloud/usearch/blob/f569c3bcbc3306fc383ef3796800ee350595b998/include/usearch/index.hpp#L2397 Change to:
```
offsets[static_cast<std::size_t>(0)] = offset + sizeof(header) + sizeof(level_t) * header.size;
```


ashvardanian:
@dranger003 there must be a way to avoid all those temporary config files and generate them on the fly in CI. We are very devoted to the idea of keeping the GIT history clean and short 😉

dranger003:
Apologies @ashvardanian, I may have missed something, can you confirm we are referring to the same files?  
There should be nine files in total:
```
csharp
|- libusearch
|  |- libusearch.vcxproj              ; native library project
|  |- libusearch.x64.nuspec           ; nuget x64 package
|  |- libusearch.x86.nuspec           ; nuget x86 package
|  \- libusearch.props                ; required for nuget (emtpy)
|- LibUSearch.Interop
|  |- LibUSearch.Interop.csproj       ; interop library project
|  \- LibUSearch.Interop.cs           ; interop library source
|- LibUSearch.Tests
|  |- LibUSearch.Tests.csproj         ; interop tests project
|  |- LibUSearch.Tests.cs             ; interop tests source
\- LibUSearch.Interop.sln             ; solution workspace
```
You can see how it looks in my repo at https://github.com/dranger003/usearch/tree/nuget/csharp.

ashvardanian:
Nothing to be sorry for @dranger003, you are a great contributor to have 🤗 I am just a complete newbie in the C# world :) 

Let me sync up with the team tomorrow on what is the absolute minimal set of ideally immutable files we need for NuGet to work - I’ll get back to you soon

dranger003:
> Nothing to be sorry for @dranger003, you are a great contributor to have 🤗 I am just a complete newbie in the C# world :)
> 
> Let me sync up with the team tomorrow on what is the absolute minimal set of ideally immutable files we need for NuGet to work - I’ll get back to you soon

Sure thing, nothing to worry! I think what you may want is to first automate compiling the native library for each platform you wish to support (i.e. also generate your company's `nuspec` and `props` files statically). Then we can remove the `libusearch` folder and the interop will continue to generate its nuget package automatically upon building.

Please note the bindings require the native library to run, and these files here give you the complete solution to generate packages that work end to end. You will need to decide with your team how to best handle your pipeline to generate packages including the interop dependency on the native library.

Feel free to pick and choose what you need and close this request. I don't recommend including nuget package files in repos, but it was requested here. Also, you typically need to sign your packages before posting them, so keep this in mind as well.

PR #197, opened 2023-08-10
Refactor of C# Bindings

AleksandrKent:
This pull request makes important updates to the C# bindings for the USearch library, mainly by adding the USearchIndex class as a user-friendly interface. 

The feature branch is rebased on top of recent state of `main-dev` branch.

### Main Changes

* New USearchIndex Class: This new class resembles the Index class in the Python code, allowing for a more unified use across different languages.

### Acknowledgment

I want to give a special thanks to @dranger003. Even though I'm changing files in the C# directory with this request, your original code was the starting point for these changes. If you find this pull request useful and have time to look at it, I would love to hear your thoughts on what could be improved or changed. I appreciate any objections and help, so please don't hesitate to share your feedback.

### Next Steps
* Add more tests with auto-generated vectors and for index save, load operations
* Auto-build native libraries and add pack with NuGet
* Feature of passing the usearch_metric_t function pointer


AleksandrKent:
@ashvardanian, I have made several updates to the build scripts and the code in the latest commits. Here are the details:

### Changes:
1. **.NET Standard 2.0 Support:** I refactored (downgraded) the code to support .NET Standard 2.0 and that limited the C# version to 7.2. As a result, I had to remove support for the `Half` (float16) type since it's not supported in C# 7.2. We have two options:
   a. Specify a higher C# version.
   b. Use `ushort[]` in the interface and provide a helper class for casting between float and ushort, if that approach is suitable.
2. **Build and Test Scripts:** The `build_and_test.cmd` and `build_and_test.sh` scripts can now be run locally. They are also used by `csharp-build-and-test.yml`, which is currently enabled only for my fork.
3. **Makefile Adjustment:** I added PHONY to the `c/Makefile` for `libusearch_c.dylib` and `libusearch_c.dll`.
4. **GitHub Actions:** The `.yml` GitHub Actions successfully pass tests on Linux and MacOS latest, but fail on Windows due to an error about a missing DLL. I am investigating this issue.
5. **Editor Configuration:** Following the guidelines from the authors of SK, I added a `.editorconfig` file where i left only settings to C# files. So it shouldn't cause troubles.

### TODOs:
1. **Windows Tests:** Investigate and fix the issue causing the tests to fail on Windows.
2. **Float16 (Half) Type:** Decide whether to specify a higher C# version or replace the type with something like `ushort`.
3. **Packaging**: The setup for packaging to NuGet is still underway. I've encountered difficulty in finding clear instructions on best practices for resolving native library paths and properly packaging them. Additional time is needed for experimentation and finding the optimal solution.

Please review these changes and let me know if you have any questions or need further adjustments.

csharp/src/LibUSearch/USearchIndex.cs
```diff
@@ -0,0 +1,262 @@
...
+        foundKeyDistances = new Dictionary<ulong, float>();
+        for (ulong i = 0; i < matches; i++)
+        {
+            foundKeyDistances.Add(keys[i], distances[i]);
+        }
+
+        return matches;
+    }
+
+    public ulong Search(float[] queryVector, ulong resultsLimit, out Dictionary<ulong, float> foundKeyDistances)
```
ashvardanian, line 208: Is it a good idea to export results into a dictionary? The order of entries will be lost, right?

csharp/src/LibUSearch/USearchIndex.cs
```diff
@@ -0,0 +1,262 @@
...
+using static NativeMethods;
+
+public class USearchIndex : IDisposable
+{
+    private IntPtr _index;
+    private bool _disposedValue = false;
+    private ulong _cached_dimensions;
+
+    public USearchIndex(
+        usearch_metric_kind_t metricKind,
```
ashvardanian, line 13: Does this enum type name match C# conventions? Looks a bit weird :)

ashvardanian (approved):
In regards to `half`-precision support, it would be great to ask the most likely first users - the Semantic Kernel team. @dluc, @shawncal, @adrianwyatt, please let us know what makes more sense - supporting legacy versions or having an `f16` interface? Is there way to have both?

csharp/src/LibUSearch/USearchIndex.cs
AleksandrKent: I became confused by the guidelines in the official documents. I recently made a commit in which I refactored the user-visible names in `USearchIndex` to match the camelCase style used in C#.

csharp/src/LibUSearch/USearchIndex.cs
AleksandrKent: No way it is a good! Thanks for pointing that out. I recently added a commit where I fixed that.

dluc:
> In regards to `half`-precision support, it would be great to ask the most likely first users - the Semantic Kernel team. @dluc, @shawncal, @adrianwyatt, please let us know what makes more sense - supporting legacy versions or having an `f16` interface? Is there way to have both?

From a SK pov both will be fine so I would suggest the most modern and less expensive approach for you, e.g. don't worry about maintaining a legacy thing. Both float and double are ok, we can easily cast with negligible cost. In scenarios where that cost would be a concern (e.g. IoT devices), the plan will be pointing users at C/C++

csharp/src/LibUSearch/USearchIndex.cs
dluc: @ashvardanian @AleksandrKent  just in case, I could run the C# code through our usual code style checker and send a PR if anything needed. The code style is a mix of .editorconfig rules and resharper settings. .editorconfig settings can be easily tested from CLI with `dotnet format`, while resharper checks are easier with a manual check just once, opening Rider (or you could ignore resharper)

Issue #206, opened 2023-08-20
Feature: Progress tracking and conditional termination from Python


ashvardanian:
### Describe what you are looking for

Most long-running tasks in the C++ layer have an optional `progress` callback argument that can report the progress and potentially terminate the job somewhere in the middle. Python and other language bindings don't currently use that feature, externally slicing the `np.ndarray` of tasks and updating `tqdm` indicator after every batch.

That, however, does not apply to `cluster`, `join`, `save`, and `load`. So we need to pass some functions from Python to C++ bindings, that will be invoked only from one thread to avoid GIL issues.

### Can you contribute to the implementation?

- [x] I can contribute

### Is your feature request specific to a certain interface?

Python bindings

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #211, opened 2023-08-23
Feature: Progress tracking and conditional termination from Python #206

mgevor:


include/usearch/index.hpp
```diff
@@ -2721,8 +2731,13 @@ class index_gt {
         using slot_level_allocator_t = typename dynamic_allocator_traits_t::template rebind_alloc<slot_level_t>;
         buffer_gt<slot_level_t, slot_level_allocator_t> slots_and_levels(size());
 
+        // Progress status
+        std::atomic<bool> do_tasks{true};
```
ashvardanian, line 2735: It's better to store the atomic error in an `std::atomic<char const *>`, like in other palces.

include/usearch/index.hpp
```diff
@@ -3411,6 +3441,9 @@ static join_result_t join(               //
     std::atomic<std::size_t> visited_members{0};
     std::atomic<char const*> atomic_error{nullptr};
 
+    // Progress status
+    std::atomic<bool> do_tasks{true};
```
ashvardanian, line 3445: No need for one more atomic variable.

include/usearch/index.hpp
```diff
@@ -3431,7 +3464,10 @@ static join_result_t join(               //
                 passed_rounds = ++rounds;
                 total_rounds = passed_rounds + free_men.size();
             }
-            progress(passed_rounds, total_rounds);
+            if (thread_idx == 0)
+                do_tasks = progress(passed_rounds, total_rounds);
+            if (do_tasks.load())
```
ashvardanian, line 3469: Isn't it the opposite condition?

ashvardanian (changes requested):

include/usearch/index.hpp
mgevor: This function doesn't return anything, so there is object to assign the error

ashvardanian:
I believe the `index.py` file must be affected by these changes, but I can't see it among the changes.
I expect `tqdm` to receive progress reports from the C++ level. In that case - `batch_size` argument for `add` and `search` will become irrelevant at the Python level, and only `log` argument will remain. @mgevor, please update accordingly.

ashvardanian (approved):

PR #213, opened 2023-08-24
160 feature c bindings and semantic kernel integration

AleksandrKent:
@ashvardanian, I've submitted a new pull request and released a NuGet package on the test servers.

**Changes**
-  Added three .yml workflows: `libs-build.yml`, `csharp-release.yml`, and `csharp-build-and-test.yml`. All workflows ran successfully prior to merging changes from main-dev. See notes below.
- Updated the `update_version.sh` to modify the C# project version.
- Renamed projects within the code.

**Notes**
- The new NuGet package is available for testing at https://int.nugettest.org/packages/USearch.NET
- I've marked future changes for production as `TODOs` in the .yml files.

**Issues**
- There's an issue with the C# tests failing across all platforms for libraries built from main-dev. The errors stem from the USearch libraries. Interestingly, some tests do pass on Windows. Everything works fine when using older versions from early August (>=1.0.0). 
**Updated**: The usearch.h API has changed. I'll update the code soon.

**Next Steps:**
- Add support for custom metrics.
- Create a script for building the USearch library on Windows.

Should we also consider adding tests for x86 architectures, or focus mainly on x64?"

Let me know your thoughts.


ashvardanian:
@AleksandrKent, we generally prefer fewer actions, as GitHub is a little slow in running them. Can we consolidate everything into `release.yml` and `prerelease.yml`? Have you tried uploading the NuGet package to the GitHub registry like we do with Maven? Is there a difference from users perspective?

AleksandrKent:
@ashvardanian 

**New Changes:**
1. Added a 'Build and Test' jobs in the prerelease.
2. Added a 'Push to NuGet' jobs for releases.
3. Verified the workflow locally.
4. For additional changes, see the commit messages.

**Notes:**
- The version is manually read from the `VERSION` file.
- The `NUGET_APIKEY` secret must be set.

**Regarding GitHub Package Registry:**
Publishing to the GitHub Package Registry is an option. However, this source is not the default for the `dotnet nuget` command. Users would need to manually add this source to their projects. Additionally, the package won't appear on nuget.org. 

We could publish to both NuGet and GitHub if it serves project needs.

AleksandrKent:
I haven't checked out the Maven part yet. I'll look into possible ways of connecting NuGet and GitHub Registry later.

Issue #207, opened 2023-08-21
Invalid: C must work with synthetic data


ashvardanian:
### Describe the bug

All tests across all languages don't depend on any external content. The same should be done for C to simplify its usage.

### Steps to reproduce

-

### Expected behavior

-

### USearch version

v1.1.1

### Operating System

Ubuntu 22.04

### Hardware architecture

x86

### Which interface are you using?

Other bindings

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #218, opened 2023-08-24
Invalid: C must work with synthetic data #207

mgevor:


mgevor:
Also, fixed memory leaks.

ashvardanian:
Would it be a good idea to add this test to GitHub CI, at least on Ubuntu?

I am also wondering if we can make cpp-check, mentioned in `compilation.md`, or some other C++ linter part of the pipeline? That may be a stricter way of checking standard compliance than `-pedantic` builds.

mgevor:
In fact, we are testing the C interface when running GoLang tests, which are based on the C interface.
But anyway, this is not explicit so that I will add it

mgevor:
Added a new test case for #222

ashvardanian (approved):

PR #221, opened 2023-08-24
Refactor: Dont set version for sphinx

gurgenyegoryan:
Sphinx 7.1.2 version installed sphinx-js-1.0, which gave unknown directive type.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Release
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


ashvardanian:
If we don't set it, we can get the errors we got in Stringzilla, if I am not mistakes. Can't we just set a specific version of the JS extension, @gurgenyegoryan ?

gurgenyegoryan:
Then I will close the PR and try to find another solution.

PR #228, opened 2023-08-27
Readme: Fix Quickstart in rust example

var77:
## Description
There are small typos in rust quick start guide, which makes assertions to fail.


ashvardanian:
@var77, thank you for your contribution! We will merge into `main` later today, with a few more improvements for our Rust users 🤗

PR #236, opened 2023-08-29
GoLang Bindings Release

davvard:


ashvardanian:
@davvard , please update the imports, return the missing `Destroy` calls, and add a section about installing the native libraries.

.github/workflows/Dockerfile.libs
```diff
@@ -0,0 +1,34 @@
+FROM ubuntu:22.04
+ENV DEBIAN_FRONTEND=noninteractive
+
+ENV user_name="runner"
+ARG TARGETPLATFORM
+ARG docker_ip
+ARG user_pass
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends python3.11 python3-pip cmake g++-12 build-essential libjemalloc-dev golang openssh-client sshpass && \
```
ashvardanian, line 10: Why do we need Python there?

.github/workflows/release.yml
```diff
@@ -96,6 +96,62 @@ jobs:
...
+        echo "docker_ip=$docker_ip" >> "$GITHUB_OUTPUT"
+
+    - name: Set new password for runner user
+      id: userpass
+      run: |
+        user_pass="1122"
+        echo "runner:$user_pass" | sudo chpasswd
+        echo "user_pass=$user_pass" >> "$GITHUB_OUTPUT"
+          
+    - name: Build Usearch lib
```
ashvardanian, line 132: How do you guess the capitalization of words? Which rule do you follow?

ashvardanian (changes requested):

PR #243, opened 2023-09-02
Fix dummy prefetch checks

al13n321:
I accidentally noticed that all the `if (!std::is_same<prefetch_at, dummy_prefetch_t>::value)` checks aren't actually working. The `prefetch_at` template argument gets inferred as `dummy_prefetch_t &`, so the `is_same` returns false, and the `prefetch()` call happens even for dummy prefetch. (Then presumably gets inlined and optimized out.)

This PR replaces the check with `if (!std::is_same<typename std::decay<prefetch_at>::type, dummy_prefetch_t>::value)`. Idk if it's the most concise or "correct" way to do it, but it works.


ashvardanian:
Hey, @al13n321! You are right, thank you for patch! I’ll mege into development branch and release soon.

PR #279, opened 2023-09-27
Faster Exact Search for C, Python, and JavaScript

ashvardanian:


ashvardanian:
This is worth merging as is. The JS part will come with a major rewrite of the binding 🤗

PR #286, opened 2023-10-11
[rust] add Sync,Send impl for Index

weedge:


ashvardanian:
Hey, @weedge! Thanks for yet another commit! Can you please explain what it does?

weedge:
@ashvardanian ,  there are some scenes, to use a global concurrent hashmap(mutex/rwlock) to store k/v (index_name/Index), view c++ impl is already threads safe, so add Sync, Send for Index in rust threads safe.  find the cxx issuecomment  like this :   https://github.com/dtolnay/cxx/issues/1175#issuecomment-1430078152

PR #291, opened 2023-10-18
[rust] add remove test

weedge:


ashvardanian:
@weedge, if you use smaller keys, the test passes. Looking into it now.

Issue #296, opened 2023-11-03
Feature: More examples in C++ docs (e.g: clustering)


deszip:
### Describe what you are looking for

For me as a developer with a slightly different tech stack it's hard to implement some features I know usearch can do for sure but doesn't expose to bindings I use (specifically swift/objc). 

Usearch has a great docs for lib itself and for different wrappers also. What I think about is to make them more consistent by covering most of features with docs and examples in c++ bc it's universal and all other wrappers use c++ API.

Feature Im trying to build is similarity grouping of photos on iOS. What Im looking for right now is an example of clustering vectors where I can get list of clusters after adding vectors to index. I can see such example in Python docs but unfortunately can't get through the implementation of the Python bindings to build the same in objc :( 

Would be nice to have this covered in c++ docs.

### Can you contribute to the implementation?

- [ ] I can contribute

### Is your feature request specific to a certain interface?

C++ implementation

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #299, opened 2023-11-05
Docs: Missing Python import

sandorkonya:
fix import error


ashvardanian:
Thank you for the patch, @sandorkonya! I will gather a few minor documentation updates and will merge them together with #296 🤗

PR #309, opened 2023-11-11
Improve: TypeScript annotations

jlarmstrongiv:
Referenced 
- Generating types from JSDoc https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html
- JSDoc object types https://stackoverflow.com/a/31573441

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -432,6 +432,7 @@ jobs:
           npm install
           npm ci
           npm test
+          npm run typegen
 
       - name: Publish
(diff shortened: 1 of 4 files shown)


jlarmstrongiv:
@ashvardanian I made a mistake in the typings—all of the `dimensionsOrConfigs` object properties are required, when only `dimensions` needs to be required.  Do you know how to make them optional with JSDoc?

Other than that one quirk, the types work great 👍

ashvardanian:
@jlarmstrongiv, sadly not familiar with that.

PR #324, opened 2023-12-26
Add libpthread for older linux gcc builds

Ngalstyan4:
Latest versions of libc (>= 2.34, released on August, 2021) have
libpthread built in but in slightly older libcs one has to still
specify -llibpthread to explicitly link it in
The commit makes sure usearch compiles successfully with these
older libcs

It changes nothing on newer libcs as libpthread exists on newer libcs as an empty library


ashvardanian:
Thank you, @Ngalstyan4! Merged and expected to be on main on Thursday 🤗

Issue #328, opened 2024-01-05
Bug: wasm packages < 400 bytes


sroussey:
### Describe the bug

The WASM packages are not valid

### Steps to reproduce

Just look at https://github.com/unum-cloud/usearch/releases

### Expected behavior

I noticed that the releases for MacOS are not correct either, so looking into both.

### USearch version

v2.8.14

### Operating System

all

### Hardware architecture

Arm

### Which interface are you using?

Other bindings

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ashvardanian:
Will patch this soon as well 🤗

komal-rs:
any update on this ?

PR #331, opened 2024-01-06
WIP: WASM builds

sroussey:
WASM is arch independant, so no need to make for each arch. Will just build on linux.

Get mmap emulation working.

We need stdlib

To use like a library requires some __attribute__ stuff in cpp and -mexec-model=reactor in the build command.

Still fails:
```
wasm-ld: error: /var/folders/74/djsf8lv965l2tkgr29ybpz4h0000gn/T/lib-a51bfe.o: undefined symbol: __cxa_allocate_exception
wasm-ld: error: /var/folders/74/djsf8lv965l2tkgr29ybpz4h0000gn/T/lib-a51bfe.o: undefined symbol: __cxa_throw
wasm-ld: error: /var/folders/74/djsf8lv965l2tkgr29ybpz4h0000gn/T/lib-a51bfe.o: undefined symbol: __cxa_free_exception
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -184,7 +184,7 @@ jobs:
         run: |
           cmake -DCMAKE_BUILD_TYPE=Release -DUSEARCH_BUILD_LIB_C=1 -DUSEARCH_BUILD_TEST_CPP=0 -DUSEARCH_BUILD_BENCH_CPP=0 -B ./build_release
           cmake --build ./build_release --config Release        
-          zip -r usearch_macOS_${{ matrix.arch }}_${{ steps.version.outputs.version }}.zip build_release/libusearch_c.so c/usearch.h
+          zip -r usearch_macOS_${{ matrix.arch }}_${{ steps.version.outputs.version }}.zip build_release/libusearch_c.* c/usearch.h
 
       - name: Upload archive
         uses: xresloader/upload-to-github-release@v1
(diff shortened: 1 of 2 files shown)


sroussey:
Related: https://github.com/unum-cloud/usearch/issues/328

ashvardanian:
I start thinking that for WASM support, we must change at least JS and Rust bindings to use the intermediate C layer, introducing ABI stability, and then simply shipping precompiled shared libraries for WASM. What do you think, @sroussey?

sroussey:
Can't ship anything that uses exceptions at the moment, at least with wasi. Quickly changing area, but still not there yet.

ashvardanian:
I am not sure about JavaScript WASM support yet, but CloudFlare recently posted an article about their Pyodide workers:

---

[LLVM](https://llvm.org/) provides three target triples for WebAssembly:

- wasm32-unknown-unknown – this backend provides no C standard library or system call interface; to support this backend, we would need to manually rewrite every system or library call to make use of imports we would define ourselves in the runtime.
- wasm32-wasi – WASI is a standardized system interface, and defines a standard set of imports that are implemented in WASI runtimes such as [wasmtime](https://github.com/bytecodealliance/wasmtime/).
- wasm32-unknown-emscripten – Like WASI, Emscripten defines the imports that a WebAssembly program needs to execute, but also outputs an accompanying JavaScript library that implements these imported functions.

Pyodide uses Emscripten, and provides three things:

- A distribution of the CPython interpreter, compiled using Emscripten
- A foreign function interface (FFI) between Python and JavaScript
- A set of third-party Python packages, compiled using Emscripten’s compiler to WebAssembly.

---

I am thinking to update the `setup.py` to support Emscripten and other compilers, and later upload the wheel to release notes.

sridhar-mani:
How to use this libraries wasm bundle.Because when I try to install with wasmer or run it gives errors.

PR #339, opened 2024-01-14
Fix some miscelenious issues encountered while working on #335

Ngalstyan4:
Fix some miscelenious issues encountered while working on #335

1. [Add move construction tests and fix an issue caused by them](https://github.com/unum-cloud/usearch/commit/78ea641c5d4a41341efb0cf2544e557f7619aa28)
ASAN flagged the memset call on empty buffer when the test ran with the added move construction portion

2. [Only consider zero length IO an error if input buffer was larger than zero](https://github.com/unum-cloud/usearch/commit/aafff91bf52e4e151e0b1487724113d38aa4729a)

5. [Move option-override policy opt-in before policy definitions so overr…](https://github.com/unum-cloud/usearch/commit/3731149c9d7d680181addd2bc283c513a9272b11)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,11 @@ if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
     set(USEARCH_IS_MAIN_PROJECT ON)
 endif ()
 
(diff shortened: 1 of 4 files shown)


ashvardanian:
Great fixes, @Ngalstyan4! Also, thank you a lot for organizing the suggestions in separate PRs, always supported by evidence! Always joy reading them 🤗

PR #342, opened 2024-01-22
Redo the javascript version and use pre-built binaries

sroussey:
This takes what I did for Simsimd and applies it to usearch.

Note the test where I changed something to be a close equals. usearch did not return 0 for that test but 1e-7 instead. Let me know if that is ok.


.github/workflows/release.yml
```diff
@@ -403,49 +398,111 @@ jobs:
...
+          access: public
 
   publish_rust:
     name: Publish Rust
     needs: versioning
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v3
-        with:
-          ref: "main"
```
ashvardanian, line 448: Without those references the old `VERSION` file is fetched and uploaded artifacts are numbered incorrectly. Can you bring them back across the file, please?

.github/workflows/prerelease.yml
```diff
@@ -2,9 +2,9 @@ name: Pre-Release
 
 on:
   push:
-    branches: ["main-dev"]
+    branches: ["main-dev", "test-ga-*"]
   pull_request:
-    branches: ["main-dev"]
+    branches: ["main-dev", "test-ga-*"]
```
ashvardanian, line 7: These are probably not needed, right?

.github/workflows/prerelease.yml
```diff
@@ -107,15 +104,14 @@ jobs:
 
   test_ubuntu_clang:
     name: Ubuntu (Clang 16)
+    if: github.ref == 'refs/heads/main'
```
ashvardanian, line 107: What is the logic here?

ashvardanian (changes requested):

.github/workflows/prerelease.yml
sroussey: Avoiding things not running on a test branch

.github/workflows/release.yml
sroussey: Honestly, I have no idea what you are referencing. The top of the file says: `branches: ["main"]` so I don't know how it would checkout any other branch. At any rate, I reverted these.

PR #346, opened 2024-01-30
Narek/misc

Ngalstyan4:
Fixed some miscelenious build issues that came up while working on lanterndata/lantern#262

See detailed comments in commit messages for each fix

Also see related changes PR in ashvardanian/SimSIMD#78

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,18 +63,11 @@ else ()
 endif ()
 
 # Core compilation settings affecting "index.hpp"
-target_compile_definitions(
-    ${USEARCH_TARGET_NAME} INTERFACE $<$<NOT:$<BOOL:${USEARCH_USE_OPENMP}>>:USEARCH_USE_OPENMP=0>
-)
+target_compile_definitions(${USEARCH_TARGET_NAME} INTERFACE "USEARCH_USE_OPENMP=$<BOOL:${USEARCH_USE_OPENMP}>")
 
 # Supplementary compilation settings affecting "index_plugins.hpp"
-target_compile_definitions(
-    ${USEARCH_TARGET_NAME} INTERFACE $<$<NOT:$<BOOL:${USEARCH_USE_FP16LIB}>>:USEARCH_USE_FP16LIB=1>
-)
-
-target_compile_definitions(
-    ${USEARCH_TARGET_NAME} INTERFACE $<$<NOT:$<BOOL:${USEARCH_USE_SIMSIMD}>>:USEARCH_USE_SIMSIMD=0>
-)
+target_compile_definitions(${USEARCH_TARGET_NAME} INTERFACE "USEARCH_USE_FP16LIB=$<BOOL:${USEARCH_USE_FP16LIB}>")
+target_compile_definitions(${USEARCH_TARGET_NAME} INTERFACE "USEARCH_USE_SIMSIMD=$<BOOL:${USEARCH_USE_SIMSIMD}>")
 
 target_include_directories(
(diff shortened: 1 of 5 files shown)


Ngalstyan4:
Updated submodule to point the latest release

include/usearch/index_plugins.hpp
```diff
@@ -50,8 +50,10 @@
 #define SIMSIMD_NATIVE_F16 !USEARCH_USE_FP16LIB
 
 #if !defined(SIMSIMD_TARGET_X86_AVX512) && defined(USEARCH_DEFINED_LINUX)
+#if defined(__AVX512F__) && defined(__AVX512FP16__) && defined(__AVX512VNNI__) && defined(__AVX512VPOPCNTDQ__)
 #define SIMSIMD_TARGET_X86_AVX512 1
 #endif
+#endif
```
ashvardanian, line 56: These should be patched in SimSIMD, not USearch, I believe. The USearch package shouldn't know exactly which subsets of AVX512 are used by SimSIMD.

c/CMakeLists.txt
```diff
@@ -20,6 +20,10 @@ if (USEARCH_BUILD_LIB_C)
     endif ()
 
     setup_target(usearch_c)
+    # Note: -static-libstdc++ is only meaningful for dynamic builds
+    # to bundle up the c++ standard library, we need to explicitly link it as below
+    target_link_libraries(usearch_static_c PRIVATE "libstdc++.a")
+
```
ashvardanian, line 26: I believe this may fail to compile on some Linux distros and with other compilers. Probably worth adding a CI pipeline for that.

CMakeLists.txt
```diff
@@ -296,14 +289,12 @@ function (setup_target TARGET_NAME)
...
+    target_compile_definitions(${TARGET_NAME} PRIVATE "USEARCH_USE_OPENMP=$<BOOL:${USEARCH_USE_OPENMP}>")
 
     # Supplementary compilation settings affecting "index_plugins.hpp"
-    target_compile_definitions(
-        ${TARGET_NAME} PRIVATE $<$<NOT:$<BOOL:${USEARCH_USE_FP16LIB}>>:USEARCH_USE_FP16LIB=1>
-    )
+    target_compile_definitions(${TARGET_NAME} PRIVATE "USEARCH_USE_FP16LIB=$<BOOL:${USEARCH_USE_FP16LIB}>")
+    target_compile_definitions(${TARGET_NAME} PRIVATE "USEARCH_USE_SIMSIMD=$<BOOL:${USEARCH_USE_SIMSIMD}>")
 
-    target_compile_definitions(${TARGET_NAME} PRIVATE $<$<NOT:$<BOOL:${USEARCH_USE_SIMSIMD}>>:USEARCH_USE_SIMSIMD=0>)
```
ashvardanian, line 306: These don't seem to be needed. The primary logic for inferring those settings is located in C/C++ sources. The CMake flags are only there to override them.

ashvardanian (changes requested):

CMakeLists.txt
Ngalstyan4: right, inference works fine. but with the old setup explicit values passed via `cmake` are ignored and these fix the issue so variables are no longer ignored

c/CMakeLists.txt
Ngalstyan4: could you elaborate on what would make this fail?

ashvardanian (approved):

ashvardanian:
Squashed and merged in 892c2b99ac55f8967488946a3706c7ddad945e9a

Issue #345, opened 2024-01-29
Bug: C# Loading usearch file exception


ErwanL08:
### Describe the bug

Hi first thing this lib is awesome thank you for this work . 

I writting an application console that embedded a list of sentence with fasttext. the result give me a FLoat[] of 300D. 
At the end i m saving the result in a specific path.
![image](https://github.com/unum-cloud/usearch/assets/16559628/692c30f2-72f5-4544-b11d-8c77fe9751b8)
If i do a search on the index it s working fine.

In a other function i try to loaded the saved index model and launch a search on it : 
![image](https://github.com/unum-cloud/usearch/assets/16559628/bdf73467-65f1-47f8-8042-374dd3504a6f)
and i have this exception : 
![image](https://github.com/unum-cloud/usearch/assets/16559628/bd2fefec-bb56-4442-84a0-d2a51821f4f8)

Do i miss something ? 

### Steps to reproduce

![image](https://github.com/unum-cloud/usearch/assets/16559628/692c30f2-72f5-4544-b11d-8c77fe9751b8)
![image](https://github.com/unum-cloud/usearch/assets/16559628/bdf73467-65f1-47f8-8042-374dd3504a6f)

### Expected behavior

no exception

### USearch version

2.8.16

### Operating System

windows 11

### Hardware architecture

x86

### Which interface are you using?

Other bindings

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ashvardanian:
Interesting. @ErwanL08 and @gebirgslok can you suggest a Pull Request extending the tests with a minimal case that can be used to reproduced the bug? I can try patching it then.

gebirgslok:
> Interesting. @ErwanL08 and @gebirgslok can you suggest a Pull Request extending the tests with a minimal case that can be used to reproduced the bug? I can try patching it then.

Sure, i'll give it a try.

BR
Jens

Geekyvk5:
Same issue, works until the save, but while loading the saved while throws an exception in C#.

ashvardanian:
@Geekyvk5 , @gebirgslok , and @ErwanL08, I've just pushed v2.9.1 which seems to fix the issue. Please let me know if you still encounter it 🤗

ErwanL08:
hi @ashvardanian thank you the bug is fix 👍

PR #350, opened 2024-02-22
add UnitTest LoadingVector and CreateFileVector

ErwanL08:
Add Unit Test for issue : 
#345

### DIFF
--- a/csharp/src/Cloud.Unum.USearch.Tests/USearchIndexTests.cs
+++ b/csharp/src/Cloud.Unum.USearch.Tests/USearchIndexTests.cs
@@ -1,3 +1,4 @@
+﻿using System.Diagnostics;
 using System.Numerics;
 using System.Runtime.InteropServices;
 using Cloud.Unum.USearch;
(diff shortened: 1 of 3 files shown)


csharp/src/Cloud.Unum.USearch.Tests/USearchIndexTests.cs
```diff
@@ -39,6 +40,54 @@ public void IndexOptions_InitializesEmptyArguments_ConstructedWithMinimalOrUnkno
...
+        using var index = new USearchIndex(savedPath);
+
+        var vector = new float[] { 0.2f, 0.6f, 0.4f };
+        int matches = index.Search(vector, 10, out ulong[] keys, out float[] distances);
+        Trace.Assert(index.Size() == 1);
+        Trace.Assert(matches == 1);
+        Trace.Assert(keys[0] == 42);
+        Trace.Assert(distances[0] <= 0.001f);
+    }
+
```
ashvardanian, line 90: For future, it is generally a bad idea to separate logic between tests. If possible, it's better to serialize and desterilize within a single function, not leaving persisted artifacts after the test ens.

PR #385, opened 2024-04-05
fix ci failure on js docs

sroussey:
I was able to test that it does not fail, but not sure if the docs are really generated:

https://github.com/sroussey/usearch/actions/runs/8565020964/job/23472451179

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -778,6 +778,10 @@ jobs:
           npm install -g jsdoc typedoc
       - name: Install USearch from PyPi
         run: pip install usearch
+      - name: Build CJS JS
+        run: |
+          npm ci --ignore-scripts
+          npm run build-js
       - name: Build documentation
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -778,6 +778,10 @@ jobs:
           npm install -g jsdoc typedoc
       - name: Install USearch from PyPi
         run: pip install usearch
+      - name: Build CJS JS
+        run: |
+          npm ci --ignore-scripts
+          npm run build-js
```
ashvardanian, line 784: Seems like we need this, but not the rest. The rest should probably be reverted 🤔

ashvardanian (changes requested):

.github/workflows/release.yml
sroussey: This makes CJS versions that the doxygen seems to expect. Then the other part points to that version.

PR #386, opened 2024-04-05
Do not use `#pragma region` if not supported by the compiler

mbautin:


ashvardanian:
Thanks for the PR! Just to clarify, which compilers don't support that pragma?

ashvardanian:
Yep, it is, but I believe there is a cleaner way to avoid this warning. You can set a push-options pragma on top of the file, and a pragma-pop, asking to ignore "unknown pragmas". That may be a less repetitive solution, assuming that region annotation is also used in SimSIMD and StringZilla 🤷‍♂️

ashvardanian:
It's [supported on GCC 13 and newer](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85487), so I'll patch the condition and merge 🤗 

There are only 4 regions in the `usearch/index.hpp`, but if we include the submodules the number multiplies to 66. Any chance there is a cleaner way to define this pragma and backport it to older GCC versions, @mbautin?

ashvardanian:
I've tried using the `_Pragma` variant, but I believe that overcomplicates things.

```cpp
// The `#pragma region` and `#pragma endregion` are not supported by GCC 12 and older.
// But they are supported by GCC 13, all recent Clang versions, and MSVC.
#if defined(__GNUC__) && ((__GNUC__ > 12) || (__GNUC__ == 12 && __GNUC_MINOR__ >= 0))
#define USE_PRAGMA_REGION
#elif defined(__clang__) || defined(_MSC_VER)
#define USE_PRAGMA_REGION
#endif
#ifdef USE_PRAGMA_REGION
#define PRAGMA_REGION(name) _Pragma(#name)
#define PRAGMA_ENDREGION(name) _Pragma(#name)
#else
#define PRAGMA_REGION(name)
#define PRAGMA_ENDREGION(name)
#endif
```

Will revert to your variant, @mbautin 🤗

PR #407, opened 2024-05-14
Extend java API with vector access

mccullocht:
Add `Index.get(int key)` method that returns the contents of the vector on the java heap. This method will throw an
`IllegalArgumentException` if key is not found in the index.

### DIFF
--- a/java/cloud/unum/usearch/Index.java
+++ b/java/cloud/unum/usearch/Index.java
@@ -123,6 +123,17 @@ public class Index {
   }
 
   /**
+   * Return the contents of the vector at key.
+   * 
+   * @param key key to lookup.
+   * @return the contents of the vector.
+   * @throws {@link IllegalArgumentException} is key is not available.
+   */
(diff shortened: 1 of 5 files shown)


java/cloud/unum/usearch/Index.java
```diff
@@ -296,7 +307,11 @@ public Config expansion_search(long _expansion_search) {
       System.loadLibrary("usearch"); // used for tests. This library in classpath only
     } catch (UnsatisfiedLinkError e) {
       try {
-        NativeUtils.loadLibraryFromJar("/usearch/libusearch.so");
+        if (System.getProperty("os.name").equals("Mac OS X")) {
```
mccullocht, line 310: This and the gradle changes were made to make it easier to develop on mac os. I've put these in a separate commit so it should be easy to drop them if desired.

ashvardanian:
Thanks for the PR! Merged 🤗

PR #410, opened 2024-05-16
Add java `Index.{load,view}FromPath(String)`

mccullocht:
If the index is `save()`ed to disk we should be able to load it without needing to either correctly guess the
parameters or hope they are ignored/overwritten when the index is loaded.

### DIFF
--- a/java/cloud/unum/usearch/Index.java
+++ b/java/cloud/unum/usearch/Index.java
@@ -46,16 +46,33 @@ public class Index implements AutoCloseable {
     long expansion_add,
     long expansion_search
   ) {
-    c_ptr =
-      c_create(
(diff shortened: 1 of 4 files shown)


ashvardanian:
Thank you for your contributions @mccullocht!

I have a Java question unrelated to this commit. Are there simpler alternatives to using Maven and SonaType. Any chance we can recommend Java developers to use Bazel or some other build tool to fetch sources from GitHub and build on-the-fly?

mccullocht:
@ashvardanian not an expert, but my impression is that most folks prefer to use precompiled artifacts rather than source distribution and bazel is unusual unless you have multiple languages in your code base.

PR #413, opened 2024-05-19
(breaking) FIX for unexpected segfault when index dimensions don't add up

jbrummack:
This should fix the problem of segfaulting by checking whether the dimensions are adding up and handling the edge case of hamming distance.
Due to `cxx::Exception` being private I had to create a new error type that wraps C++ Exceptions and can output errors from "rustland".
**Therefore this fix could introduce breaking changes** 
It works in my (small) codebase and passes tests, but due to finding more cases that cause an unexpected segfault in rust, I suggest merging these fixes together later.

### DIFF
--- a/rust/lib.rs
+++ b/rust/lib.rs
@@ -474,15 +474,6 @@ pub enum MetricFunction {
     F64Metric(std::boxed::Box<dyn Fn(*const f64, *const f64) -> Distance + Send + Sync>),
 }
 
-impl MetricFunction {
-    fn is_variable_length(&self) -> bool {
-        match self {
-            B1X8Metric => true,
-            _ => false,
-        }
-    }
-}
-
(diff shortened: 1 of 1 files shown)


ashvardanian:
If we perform the size check in the C layer - we would avoid breaking changes and could merge as is. How does that sound?

jbrummack:
Sounds like a better solution. Adding some more tests in the Rust part is probably a good idea; throwing out the changes in Rust except the test for having different vector sizes and relying on the C layer for making it pass.

ashvardanian:
Yes, @jbrummack! Please let me know if you'd have time to implement those changes 🤗

jbrummack:
I certainly have time to implement more Rust tests, as its probably synergistic with using the library in projects. 
I am not really constrained by time, but looking at the c layer would take a while since I don't have much experience with c/c++ (yet).

Issue #418, opened 2024-05-22
Bug: Syntax Error with Jest in ESM


eryue0220:
### Describe the bug

When run test through Jest in ESM, it will occur an error like:
![image](https://github.com/unum-cloud/usearch/assets/3050418/02e9bf2a-5426-45e0-a196-09c564e76df0)

### Steps to reproduce

1. git clone https://github.com/eryue0220/usearch-jest-esm-bug
2. cd usearch-jest-esm-bug
3. pnpm i
4. pnpm test

### Expected behavior

It should pass without babel transform the code.

### USearch version

2.11.1

### Operating System

Mac OSX

### Hardware architecture

Arm

### Which interface are you using?

Other bindings

### Are you open to being tagged as a contributor?

- [X] I am open to being mentioned in the project `.git` history as a contributor

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

eryue0220:
Currently, A workaround is provide a transform in jest to remove `module.exports` in source code: https://github.com/eryue0220/usearch-jest-esm-bug/blob/main/jest.config.mjs#L5

ashvardanian:
I am very inexperienced in JS tooling. Any chance you could open a PR if you know how to fix that?

eryue0220:
> I am very inexperienced in JS tooling. Any chance you could open a PR if you know how to fix that, @eryue0220?

Sure, I will fix this later.

PR #420, opened 2024-05-23
fix: esm bug in jest

eryue0220:
This pr is to solve #418

### DIFF
--- a/javascript/usearch.ts
+++ b/javascript/usearch.ts
@@ -609,15 +609,6 @@ function exactSearch(
   }
 }
 
-module.exports = {
-  Index,
-  MetricKind,
-  ScalarKind,
-  Matches,
-  BatchMatches,
-  exactSearch,
-};
-
 // utility functions to help find native builds
 
 function getBuildDir(dir: string) {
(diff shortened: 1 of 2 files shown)


ashvardanian (approved):

ashvardanian:
Can this break compatibility with other JS environments?

eryue0220:
> Can this break compatibility with other JS environments?

Oh, you're right. It will break esm compatibility 🤣, but I forgot it. Because nodejs requires the esm module must have a `.mjs` extension or  add `"type": "module"` in `package.json`. 

I think I need to upgrade the build tool to `vite`, because `tsc` couldn't configure the target file extension.

eryue0220:
Sorry, I didn't notice that `dist-package-esm.json` has done what I said before. Now it won't break compatibility with other JS environments.

ashvardanian:
@eryue0220 did you push any changes? Are the module exports supposed to be deleted?

eryue0220:
> @eryue0220 did you push any changes? Are the module exports supposed to be deleted?

No any other changes. Yes it's supposed to be deleted.

PR #428, opened 2024-05-31
Reset cast buffer so it can be set with bitwise ORs

Ngalstyan4:
When converting floating point arrays to binary, we use bitwise OR operations to set the relevant bits in the output
buffer to 1. We do nothing if the bit is zero, so we assume that the bit is zero to start with.
The `memset` statement makes sure this assumption holds.


ashvardanian:
Amazing, thank you @Ngalstyan4, merged 😉

Issue #314, opened 2023-11-17
Java JAR contains native library twice


ccjernigan:
### Describe the bug

There are two copies of the libusearch.so in the JAR file.  This shouldn't break anything, but it would cause the JAR to be larger than it should be.

shasum -a 256 libusearch.so 
0eed2c593261e727b31e83b75323de927959b4bdf36d3d48c7ce2007e018674c  libusearch.so

shasum -a 256 shared/libusearch.so
0eed2c593261e727b31e83b75323de927959b4bdf36d3d48c7ce2007e018674c  shared/libusearch.so

<img width="725" alt="image" src="https://github.com/unum-cloud/usearch/assets/699907/0a5b0c27-9019-49bd-b45e-73b1f06e8f29">

### Steps to reproduce

1. Download the 2.8.12 JAR from Maven central
2. Unzip the JAR
3. Inspect the JAR contents

### Expected behavior

Jar contains a single 

### USearch version

2.8.12

### Operating System

n/a

### Hardware architecture

Arm

### Which interface are you using?

Other bindings

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ashvardanian:
Hey @ccjernigan! Thank you for spotting the issue! Any chance you know how to fix it? I'm only the basics of using Java and Bazel.

PR #430, opened 2024-06-01
Fix: Java JAR contains native library twice

MarkReedZ:
This PR fixes https://github.com/unum-cloud/usearch/issues/314 by excluding the shared/ directory.

Tested with

```
javac -cp /usr/share/java/junit4.jar:../../build/libs/usearch-2.13.0.jar IndexTest.java
java -cp .:/usr/share/java/junit4.jar org.junit.runner.JUnitCore IndexTest
```

### DIFF
--- a/java/cloud/unum/usearch/Index.java
+++ b/java/cloud/unum/usearch/Index.java
@@ -375,9 +375,9 @@ public class Index implements AutoCloseable {
     } catch (UnsatisfiedLinkError e) {
       try {
         if (System.getProperty("os.name").equals("Mac OS X")) {
-          NativeUtils.loadLibraryFromJar("/usearch/libusearch.dylib");
+          NativeUtils.loadLibraryFromJar("/usearch/shared/libusearch.dylib");
         } else {
-          NativeUtils.loadLibraryFromJar("/usearch/libusearch.so");
+          NativeUtils.loadLibraryFromJar("/usearch/shared/libusearch.so");
         }
       } catch (IOException e1) {
         throw new RuntimeException(e1);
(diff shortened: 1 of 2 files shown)


MarkReedZ:
Note that a cleaner fix for this issue would probably be to not copy the dlls and load them from `usearch/shared/`

In the build.gradle we copy `usearch/shared/*` to `../` then in the code we load from `/usearch`.  Instead we can remove the copying and load from `/usearch/shared/`
```
          NativeUtils.loadLibraryFromJar("/usearch/libusearch.dylib");
        } else {
          NativeUtils.loadLibraryFromJar("/usearch/libusearch.so");
```

ashvardanian:
The latter sounds like a good plan. I can wait for that patch, if you want to follow up on this branch 🤗

MarkReedZ:
Proper fix and verified the jar as in the OP.

java/cloud/unum/usearch/Index.java
```diff
@@ -375,9 +375,9 @@ public Config expansion_search(long _expansion_search) {
     } catch (UnsatisfiedLinkError e) {
       try {
         if (System.getProperty("os.name").equals("Mac OS X")) {
```
ashvardanian, line 377: Do we know what this property is equal to on Windows? We should probably link to .dll there.

ashvardanian (approved):

ashvardanian:
`build.gradle:217-227` now uses a single `from(...).into("usearch-native")` block. Thanks for help, @MarkReedZ! Closing as resolved 🤗

PR #466, opened 2024-08-19
Wrap throw statements to enable building library with noexceptions

sdenton4:


sdenton4:
As discussed on discord:
Usearch's index.hpp has some macros for wrapping 'throw' statements, theoretically allowing compilation without exceptions. However, there are a few throw statements in index_dense.hpp and index_plugins.hpp which were unwrapped. This adds the wrapping.

sdenton4:
Thanks, cap'n! I was a bit slow putting it together, as I was checking that I could get the whole thing to build end-to-end (from simsimd up) in our rather particular build system. Things are now looking good and (shockingly!) I didn't hit any further roadblocks.

PR #474, opened 2024-08-25
Upgrade: update package-lock.json

abetomo:
I ran `npm install` and it automatically updated.


abetomo:
I see, we were updating with CI.

PR #486, opened 2024-09-09
Fix: JavaScript Fix bug in add() that aborts with invalid operations

abetomo:
For example, if you try to add the same key, it aborts.

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Duplicate keys not allowed in high-level wrappers
Aborted (core dumped)
```

Improved error handling to throw JavaScript exceptions.

### DIFF
--- a/javascript/lib.cpp
+++ b/javascript/lib.cpp
@@ -161,7 +161,11 @@ void CompiledIndex::Add(Napi::CallbackInfo const& ctx) {
     auto run_parallel = [&](auto vectors) {
         executor_stl_t executor;
         executor.fixed(tasks, [&](std::size_t /*thread_idx*/, std::size_t task_idx) {
-            native_->add(static_cast<default_key_t>(keys[task_idx]), vectors + task_idx * native_->dimensions());
+            try {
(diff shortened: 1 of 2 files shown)


javascript/lib.cpp
```diff
@@ -161,7 +161,11 @@ void CompiledIndex::Add(Napi::CallbackInfo const& ctx) {
     auto run_parallel = [&](auto vectors) {
         executor_stl_t executor;
         executor.fixed(tasks, [&](std::size_t /*thread_idx*/, std::size_t task_idx) {
-            native_->add(static_cast<default_key_t>(keys[task_idx]), vectors + task_idx * native_->dimensions());
+            try {
+                native_->add(static_cast<default_key_t>(keys[task_idx]), vectors + task_idx * native_->dimensions());
+            } catch (const std::runtime_error& e) {
+                Napi::Error::New(ctx.Env(), e.what()).ThrowAsJavaScriptException();
+            }
```
ashvardanian, line 168: The `add` operation returns and `add_result_t`, so we don't have to `try-catch` the exception and can check the result object directly.

ashvardanian (changes requested):

ashvardanian (approved):

PR #495, opened 2024-09-27
Fix: Android build problem

CCnut:


ashvardanian:
This is amazing, thank you, @CCnut! Any chance you could also add a CI pipeline in prerelease.yml to make sure this build always passes and we don’t break things?

PR #503, opened 2024-10-10
Clustering & Minor Patches in JS, Rust, & Java SDKs

ashvardanian:
Many data scientists embark on their journey by implementing K-Means clustering, much like app developers starting with a calculator. But despite K-Means’ popularity, most implementations overlook the power of SIMD on modern CPUs. Efficient vector math, especially with single- and double-precision floating-point vectors, is challenging due to the computational cost of accuracy. Meanwhile, `float16`, `bfloat16`, and smaller types can fail under uneven distributions or when computing centroids for large clusters. So, what’s Unum’s solution? **Mixed precision!**

Thanks to strong community support and sponsorship from @sutoiku ([LinkedIn](https://www.linkedin.com/company/sutoiku), [Website](https://stoic.com/)), we're introducing a high-performance K-Means implementation! It utilizes any numeric type for distance calculations, switching to `float64` for centroid updates, a technique that boosts performance and enables billion-scale clustering on a single machine.

PR #504, opened 2024-10-12
Test: JS: Skip view() test for now, because it fails on Winodws

abetomo:
The test itself succeeds, but fails with the following error when deleting the index file created by save() in afterEach().

```
error: "EBUSY: resource busy or locked, unlink 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\usearch.test.index'"
```

Since it is only in Winodws that it fails, we will skip it on Winodws for now.
We will continue to investigate the solution.

### DIFF
--- a/javascript/usearch.test.js
+++ b/javascript/usearch.test.js
@@ -158,7 +158,10 @@ test('Serialization', async (t) => {
         assertAlmostEqual(results.distances[0], new Float32Array([0]));
     });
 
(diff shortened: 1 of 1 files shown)


abetomo:
Maybe the file opened with view() is not closed, but I couldn't find a solution right away, so I skipped it for now.

Related: #503

PR #506, opened 2024-10-20
Fix: Fix crash when trying to remove with read-only index

abetomo:
The index read by view() is read-only.
When I did a remove() on that index, it crashed. (`Segmentation fault (core dumped)`)

Although it is an invalid operation to do a remove() on a view() index, it is not a better way to crash, so we have fixed it to return an error.


ashvardanian (approved):

abetomo:
Thanks for the review.
I fixed the test as it was failing. #509

PR #510, opened 2024-10-24
Fix: JS: Improve error handling of batch addition

abetomo:
Crash if there is an error in batch addition.
Example:

```js
const usearch = require('usearch');
const index = new usearch.Index({ metric: 'l2sq', connectivity: 16, dimensions: 3 });
index.add(42n, new Float32Array([0.2, 0.6, 0.4]));
index.add(
  [41n, 42n, 43n],
  [[0.1, 0.6, 0.4], [0.2, 0.6, 0.4], [0.3, 0.6, 0.4]]
)
```

`42n` is duplicated, which causes an error, but then it crash.

It seems that it is not better to throw an exception during the process, so we changed it to throw it after it is completed.

### DIFF
--- a/javascript/lib.cpp
+++ b/javascript/lib.cpp
@@ -173,7 +173,12 @@ void CompiledIndex::Add(Napi::CallbackInfo const& ctx) {
                                                vectors + task_idx * native_->dimensions());
             if (!result) {
                 std::lock_guard<std::mutex> lock(mtx);
-                error += "<key:" + std::to_string(keys[task_idx]) + " message:" + result.error.release() + ">";
+                error
+                    .append("<key:")
+                    .append(std::to_string(keys[task_idx]))
+                    .append(" message:")
+                    .append(result.error.release())
+                    .append(">");
             }
         });
         if (error.size() > 0)
--- a/javascript/lib.cpp
+++ b/javascript/lib.cpp
@@ -44,6 +44,7 @@ class CompiledIndex : public Napi::ObjectWrap<CompiledIndex> {
     Napi::Value Count(Napi::CallbackInfo const& ctx);
 
     std::unique_ptr<index_dense_t> native_;
+    std::mutex mtx;
 };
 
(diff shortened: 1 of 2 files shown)


abetomo:
This patch fixes the crash, but is there a better way?
I am drafting this because I would like to change it if there is a best way to fix it.

javascript/lib.cpp
```diff
@@ -165,13 +165,16 @@ void CompiledIndex::Add(Napi::CallbackInfo const& ctx) {
...
         executor_stl_t executor;
         executor.fixed(tasks, [&](std::size_t /*thread_idx*/, std::size_t task_idx) {
             add_result_t result = native_->add(static_cast<default_key_t>(keys[task_idx]),
                                                vectors + task_idx * native_->dimensions());
             if (!result)
-                Napi::Error::New(ctx.Env(), result.error.release()).ThrowAsJavaScriptException();
+                error += "<key:" + std::to_string(keys[task_idx]) + " message:" + result.error.release() + ">";
         });
+        if (error.size() > 0)
+            Napi::Error::New(ctx.Env(), error).ThrowAsJavaScriptException();
```
ashvardanian, line 177: This is not thread-safe. You need a `mutex` to safeguard the error string. Moreover, your solution may result in a heap allocation in the default string constructor. There should be a way to avoid it.

ashvardanian (changes requested):

javascript/lib.cpp
abetomo: Thanks for the review.
I understand that adding it to a string is not a good idea, but I would like to know if there is a better way to handle errors in such cases.

Would a Python implementation be helpful?
https://github.com/unum-cloud/usearch/blob/main/python/lib.cpp#L180-L211

ashvardanian:
Hi @abetomo! I've renamed the commits and merged them into `main-dev`, changing the way multi-threading and error-setting happens in the JS binding. Please let me know what you think 🤗

PR #517, opened 2024-11-01
Fix: Avoid overflow when viewing large files

rschu1ze:
ClickHouse compiles Usearch
- with `USEARCH_USE_FP16LIB` on all platforms
- with `USEARCH_USE_SIMSIMD` on x86 and ARM.

I.e. on x86 and ARM, both defines are set. Previously, the code would do
the fp32 <--> fp16 conversion using fp16lib but I think it should prefer
the "homegrown" conversion routines (SimSIMD).


ashvardanian:
The `fp16lib` seemingly handles corner cases better. Have you noticed some issues, @rschu1ze?

rschu1ze:
Didn't have issues, no need to merge this PR then. Thanks.

PR #526, opened 2024-11-03
Remove fp16

Spixmaster:
The library is not well maintained and is only needed for already covered functionality.
fp16 has no released version.


ashvardanian:
https://github.com/ashvardanian/SimSIMD/blob/fac489df4e1edbf2262c216883b7fc2d7ec47dd0/include/simsimd/types.h#L494

PR #538, opened 2024-11-21
Make: Include submodule header files

abetomo:
For use in a C++ project, we configured CMakeLists.txt to fetch and build USearch.

For example, the following configuration:
```cmake
fetchcontent_declare(
  usearch
  GIT_REPOSITORY "https://github.com/unum-cloud/usearch.git"
  GIT_TAG "v2.16.5"
  GIT_SUBMODULES ...)
fetchcontent_makeavailable(usearch)
```

Then it cannot find the header file for the submodule and an error occurs.

For example, the following errors:
```
.../_deps/usearch-src/include/usearch/index_plugins.hpp:37:10: fatal error: fp16/fp16.h: No such file or directory
   37 | #include <fp16/fp16.h>
```

So we added a header file to be included in target_include_directories().


abetomo:
Thanks for the quick response.

ashvardanian:
It was an easy call @abetomo, thanks to you!

PR #547, opened 2025-01-09
Fix Python `usearch.index.search` in non-exact similarity search

Stefano-t:
The  Python's`usearch.index.search` method creates an `Index` when performing a similarity search with `exact = False`. `Index` requires all its inputs to be passed by name, but the first parameter, the dimension, was not supplied like that.

This PR fixes this error by assigning a name to the first parameter of the `Index` instance, specifically `ndim`.

### DIFF
--- a/python/usearch/index.py
+++ b/python/usearch/index.py
@@ -1530,7 +1530,7 @@ def search(
 
     if not exact:
         index = Index(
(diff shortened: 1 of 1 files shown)


ashvardanian:
Thank you, @Stefano-t!

Issue #554, opened 2025-01-16
Feature: Swift exceptions


danpalmer:
### Describe what you are looking for

Currently the Swift bindings are just generated from Objective-C. Unfortunately Objective-C uses a different error/exception model to Swift.

Currently, errors are thrown as `NSException`s. These are impossible to catch in Swift, and therefore cause a crash.

It would be great if these were thrown as Swift Exceptions. The Swift/Obj-C translation process can automate part of this with annotations.

### Can you contribute to the implementation?

- [ ] I can contribute

### Is your feature request specific to a certain interface?

Other

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ashvardanian:
Any chance you can draft a minimal solution?

danpalmer:
The pattern in Objective-C that was widely used for errors (exceptions are closer to a panic/fatal error, but not quite), was to pass in an error pointer and have the function set that on return. This pattern was widely enough used that the Objective-C to Swift translation layer understands it with a given attribute and converts it to a thrown Swift error.

Example:

```objc
- (UInt32)filteredSearchSingle:(Float32 const *_Nonnull)vector
                 count:(UInt32)wanted
                filter:(USearchFilterFn)predicate
                  keys:(USearchKey *_Nullable)keys
             distances:(Float32 *_Nullable)distances
                 error:(NSError **)errorPtr __attribute__((swift_error(nonnull_error))){
    search_result_t result = _native->filtered_search(vector, static_cast<std::size_t>(wanted), predicate);

    if (!result) {
        *errorPtr = [NSError errorWithDomain:@"YourErrorDomain" 
                                           code:100 
                                       userInfo:@{NSLocalizedDescriptionKey: [NSString stringWithUTF8String:result.error.release()]}]; 
        return 0;
    }

    std::size_t found = result.dump_to(keys, distances);
    return static_cast<UInt32>(found);
}
```

```swift
do {
    try index.filteredSearchSingle(vec, count: 10, filter: fn, keys: nil, distances: nil)
} catch {
    println(error)
}
```

ashvardanian:
Will it still work with the older ObjC code or do we need to add new APIs to avoid breaking changes?

danpalmer:
This would be backwards-incompatible, although in my opinion a strong improvement to the ObjC API. From my (outdated?) experience, error passing was more common than NSException, and more idiomatic.

[...]

PR #555, opened 2025-01-19
Improve error handling in Objective-C and Swift bindings

danpalmer:
This is a **breaking change** to the Objective-C and Swift bindings.

**Why?** – Because `NSException` which is currently used, is designed for unrecoverable runtime exceptions, not control flow. In Objective-C this is awkward to handle, and in Swift it's impossible to catch.

This change switches from throwing `NSException`s to returning `NSError`s, in the style recommended for handling cross-language errors in the Apple documentation: https://developer.apple.com/documentation/swift/handling-cocoa-errors-in-swift. This is worth a breaking change in order to accurately represent the possibilities of errors in the API contracts in both languages.

The Objective-C APIs include an `(NSError**)error` argument, allowing callsites to pass an empty error pointer to be populated for error handling. This is a very common pattern in Objective-C APIs.

```objc
NSError *error = nil;
[index add:foo error:&error];
if (error != nil) {
    // handle error
}
```

In Swift, this is translated to a throwing error in the normal Swift way:

```swift
try {
    index.add(foo)
} catch {
    println("\(error)")
}
```

Additionally, the error enum is bridged to Swift so that errors can be disambiguated.

Fixes: #554

### DIFF
--- a/objc/USearchObjective.mm
+++ b/objc/USearchObjective.mm
@@ -14,6 +14,8 @@ using labeling_result_t = typename index_dense_t::labeling_result_t;
 using search_result_t = typename index_dense_t::search_result_t;
 using shared_index_dense_t = std::shared_ptr<index_dense_t>;
 
+NSErrorDomain const USearchErrorDomain = @"cloud.unum.usearch.USearchErrorDomain";
+
 static_assert(std::is_same<USearchKey, index_dense_t::vector_key_t>::value, "Type mismatch between Objective-C and C++");
 
 metric_kind_t to_native_metric(USearchMetric m) {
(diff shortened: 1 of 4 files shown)


objc/USearchObjective.mm
```diff
@@ -278,102 +325,134 @@ - (UInt32)searchHalf:(void const *_Nonnull)vector
...
 
-- (Boolean)contains:(USearchKey)key {
+- (Boolean)contains:(USearchKey)key error:(NSError**)error {
     return _native->contains(key);
 }
 
-- (UInt32)count:(USearchKey)key {
+- (UInt32)count:(USearchKey)key error:(NSError**)error {
     return static_cast<UInt32>(_native->count(key));
 }
```
ashvardanian, line 355: Is there no way to highlight those functions as `noexcept` in C++ terms? That why they probably still be accessed as computed properties of the class instance, right?

danpalmer: I don't know. If you think the implementation of these will _never_ throw, and don't want to reserve the ability to throw in the API then I can drop these though.

ashvardanian:
Hi @danpalmer! Thank you for your notes and PRs! I'm now merging your new Swift → C → C++ binding and I assume it closes this issue, right?

ashvardanian:
I’ve tested the merged stuff on Linux and it worked fine after a couple of minor patches. So this PR should be closed?

danpalmer:
Yep, closing, thanks!

PR #558, opened 2025-01-22
[WIP] Add bf16 support for adding to the index

egorovcharenko:


ashvardanian:
Looks good to me!

ashvardanian:
I guess we can combine this with #561. @egorovcharenko, you've mentioned that some tests are failing. Do you remember which ones?

egorovcharenko:
For me, the tests with adding and searching in an index didn’t find the bf16 value I think.

Issue #564, opened 2025-02-06
Bug: segfault when run usearch_search() concurrently with golang


cpegeric:
### Describe the bug

When run concurrently with golang api concurrently, it will segfault or with error mummap()/free(): invalid pointer.

With -DUSEARCH_USE_OPENMP=1,  usearch_search will segfault.
```
[0 1 2] [0 3 12]
    index_test.go:102: 
        	Error Trace:	/home/ubuntu/p/matrixone/pkg/vectorindex/index_test.go:102
        	            				/home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.0.linux-amd64/src/runtime/asm_amd64.s:1700
        	Error:      	Not equal: 
        	            	expected: 0x11
        	            	actual  : 0x0
        	Test:       	TestUSearch
    index_test.go:102: 
        	Error Trace:	/home/ubuntu/p/matrixone/pkg/vectorindex/index_test.go:102
        	            				/home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.0.linux-amd64/src/runtime/asm_amd64.s:1700
        	Error:      	Not equal: 
        	            	expected: 0x11
        	            	actual  : 0x0
        	Test:       	TestUSearch
SIGSEGV: segmentation violation
PC=0x7f464b97e63e m=4 sigcode=1 addr=0x11cfa00
signal arrived during cgo execution

goroutine 58 gp=0xc000129180 m=4 mp=0xc00004d508 [syscall]:
runtime.cgocall(0x5b3931, 0xc00005be28)
	/home/ubuntu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.0.linux-amd64/src/runtime/cgocall.go:167 +0x4b fp=0xc00005be00 sp=0xc00005bdc8 pc=0x46c06b
github.com/unum-cloud/usearch/golang._Cfunc_usearch_search(0xcb4350, 0xc0004cecb0, 0x1, 0x3, 0xc0004cd9c8, 0xc0004cecc0, 0xc0000ad3f0)
	_cgo_gotypes.go:548 +0x4c fp=0xc00005be28 sp=0xc00005be00 pc=0x538b6c
github.com/unum-cloud/usearch/golang.(*Index).Search.func1(0xc00005bec8?, 0xc00005bf40, 0x3, 0xc00005bf28?, 0x5b0f25?, 0xc0000ad3f0)
[...]

PR #565, opened 2025-02-07
Bug: segfault when run usearch_search() concurrently with golang #564

cpegeric:
fixed issue Bug: segfault when run usearch_search() concurrently with golang #564

code didn't check try_pop return false and continue to proceed and segfault as result.

loop try_pop until success.

### DIFF
--- a/include/usearch/index_dense.hpp
+++ b/include/usearch/index_dense.hpp
@@ -1971,7 +1971,14 @@ class index_dense_gt {
 
         available_threads_mutex_.lock();
         usearch_assert_m(available_threads_.size(), "No available threads to lock");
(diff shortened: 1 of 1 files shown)


ashvardanian:
I'm afraid it's not the right way to fix this. There is already an `assert` above that block that says, that states that you shouldn't call this function if you don't have threads available. Also, blocking busy-waiting is also generally discouraged.

A better solution would be to add one more `Reserve` method for both size and a thread count, @cpegeric 🤗

cpegeric:
the assert is not working in this case.  assert true (means thread available) but try_pop() return false (means avaialble threads is empty).

PR #566, opened 2025-02-07
Bug: segfault when run usearch_search() concurrently with golang

cpegeric:
Fixed Bug: segfault when run usearch_search() concurrently with golang #564

Add ChangeThreadsAdd() and ChangeThreadsSearch() function to golang api to change the thread limit.

Golang user should stop calling search() function when thread limit reached.

go fmt the code.


ashvardanian:
Thank you @cpegeric!

PR #575, opened 2025-03-21
Fix a correctness bug caused by a wrong memory override

ycw66:
### Related issue
#576

### Proposed Fix
In `context` struct, we can add another member(called `top_for_refine`) of type `top_candidate_t`, which cost very limited memory , actually the capacity is at most 32, because max connectivity is 32 in usearch hnsw algorithm. 

During the execution of `form_reverse_links_`, we use the new added member `top_for_refine` to execute the `refine_` function.

As we add the top_for_refine in the context, it avoid the overhead of repeated construction and deconstruction during the execution of vector index.

### DIFF
--- a/include/usearch/index.hpp
+++ b/include/usearch/index.hpp
@@ -2499,13 +2499,12 @@ class index_gt {
         if (nodes_)
             std::memcpy(new_nodes.data(), nodes_.data(), sizeof(node_t) * size());
 
+        // Pre-reserve the capacity for `top_for_refine`, which always contains at most one more
+        // element than the connectivity factors.
         std::size_t connectivity_max = (std::max)(config_.connectivity_base, config_.connectivity);
-        // pre-reserve the capacity for 'top_for_refine'
-        for (size_t i = 0; i < new_contexts.size(); ++i) {
-            if (!new_contexts[i].top_for_refine.reserve(connectivity_max + 1)) {
+        for (std::size_t i = 0; i != new_contexts.size(); ++i)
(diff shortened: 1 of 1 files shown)


ashvardanian:
Thanks for sharing the benchmarks in the #576! Very promising!

The suggested patch, however, doesn't handle potential memory allocation failure in `reserve`. Can you make sure the memory is reserved in a similar fashion to other queues?

ycw66:
> Thanks for sharing the benchmarks in the #576! Very promising!
> 
> The suggested patch, however, doesn't handle potential memory allocation failure in `reserve`. Can you make sure the memory is reserved in a similar fashion to other queues, @ycw66?

Hi @ashvardanian ! 

Thank you for your advice! I’ve considered the memory allocation issue and believe we can handle it by pre-reserving the capacity for top_for_refine. Since its maximum capacity will never exceed connectivity_max + 1.

ycw66:
> Thanks for sharing the benchmarks in the #576! Very promising!
> 
> The suggested patch, however, doesn't handle potential memory allocation failure in `reserve`. Can you make sure the memory is reserved in a similar fashion to other queues, @ycw66?

Hi @ashvardanian! I am kindly asking whether this PR should be revised again?

ashvardanian:
Hi @ycw66! Thanks for the patch!
I can see it addresses the discussed issue, but haven't had the chance to test it myself.
I'll try to check it today, but have a few other urgent things to juggle this week 😞

PR #578, opened 2025-03-31
Fix key type for Java bindings

agoncharuk:
Currently, the default native key type of dense index `index_dense_t::vector_key_t` does not match the key type declared for Java bindings. This causes two issues:
* The memcopy in `result::dump_to` overruns the allocated buffer which eventually causes malloc to fail with self-check
* The results returned by search are invalid because when long array is interpreted as int, it causes every second value to be 0 (the current test does not catch this because it checks only one result)

Additionally, this PR fixes a case when the index size is smaller than wanted and eliminates extra copy by dumping results directly to the result array.


ashvardanian:
Thank you, @agoncharuk! All tests pass, merging!

PR #579, opened 2025-04-01
Test: Add test case for failing `del index[<key>]`

IoeCmcomc:
In Python, calling `del index[<key>]` on an `Index` object always raises `TypeError: exceptions must derive from BaseException`.

This PR adds a test case to reproduce the bug.


ashvardanian:
Thank you for the PR, @IoeCmcomc 🤗

PR #588, opened 2025-04-14
Trying to prefetch before exact search

mapleFU:


ashvardanian:
Hi @mapleFU! What's the usecase for prefetching the entire dataset?

mapleFU:
Sorry for late reply, so prefetching is for prefetch before exact data item, and prefetch the whole dataset would be regarded as "can be done by underlying storage/os"?

ashvardanian:
No worries, @mapleFU! That use-case may be problematic, as prefetch doesn't return anything and thus will end up throwing from a noexcept function. We may want to postpone such workloads until a bigger redesign 🤷‍♂️

PR #596, opened 2025-04-22
Extend test suite to int8 and uint8

liquidcarbon:
Little something to get going.

There appears to be unused global `dtypes` which I've extended with `np.int8` and `np.uint8` and used in the first test.

Tests pass with int8 but not uint8.

Shall I modify the rest of the tests in a similar manner (will check for nuances)?

### DIFF
--- a/python/scripts/test_index.py
+++ b/python/scripts/test_index.py
@@ -29,7 +29,7 @@ quantizations = [
     ScalarKind.BF16,
     ScalarKind.I8,
 ]
-dtypes = [np.float32, np.float64, np.float16]
+dtypes = [np.float32, np.float64, np.float16, np.int8, np.uint8]
 threads = 2
 
 connectivity_options = [3, 13, 50, DEFAULT_CONNECTIVITY]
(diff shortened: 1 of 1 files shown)


ashvardanian:
I'll take it from here, thanks @liquidcarbon!

ashvardanian:
Hi @liquidcarbon! The issue is coming from the constraints of Python's & NumPy's type-system. There is no separate definition for bit-packs, so I've reserved the `u8` values in Python for bit arrays. So your inputs are being treated as individual bits. [In SimSIMD](https://github.com/ashvardanian/simsimd?tab=readme-ov-file#one-to-one-distance), I've addressed it by forcing users to pass an additional `dtype`-like parameter (like `"bin8"` for 8-bit packs).

I can perform a similar change in USearch, but it will break the existing behaviour. I'll tag it accordingly and get back to it for v3. Thanks!

liquidcarbon:
Maybe it's a good idea to steer clear of unsigned ints altogether.
image

PR #599, opened 2025-04-24
Fix: filtered_search should update radius when top not empty

mapleFU:


mapleFU:
gentle ping @ashvardanian

PR #600, opened 2025-04-30
Allow kwargs in Index.restore

frank-lsf:
I want to allow kwargs to `Index.restore()` so that I can pass in arguments like `enable_key_lookups=False`.


ashvardanian:
Thank you, @frank-lsf! Merged 😉

PR #606, opened 2025-05-14
Accuracy, Correctness, and JS Multi-threading

ashvardanian:
This is a major minor release 😅 
It doesn't break anything, doesn't change too much, but will be useful to all users, and is mostly driven by the community 🤗

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -151,6 +151,7 @@
     "FAISS",
     "fbin",
     "furo",
+    "geospatial",
     "googleanalytics",
     "groundtruth",
     "hashable",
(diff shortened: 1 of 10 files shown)


javascript/usearch.ts
```diff
@@ -351,8 +354,14 @@ export class Index {
       );
     }
 
+    if ((!Number.isNaN(threads) && typeof threads !== "number") || threads < 0) {
```
abetomo, line 357: Since the validation of `threads` is the same, it would be better to make it a method.
I will refactor it if you like.

abetomo (approved):

PR #607, opened 2025-05-16
Fix multithread error after index restore

Stefano-t:
This fix removes an annoyance in Python bindings when restoring an index from disk. In particular, the following code used to have the issue:

```python
index = Index.restore("/path/to/index.usearch", view=True)  # also view=False has the problem.
result = index.search(..., threads=2)  # or any number >= 2
```

The search triggered a "Can't use that many threads" error.

However, `threads=1` doesn't cause any issue, nor does `threads=0`. Furthermore, using `threads=0` and later any `threads=N` with `N>=2` resulted in a valid search.

Digging around the code, I've found that `threads=0` calls a `try_reserve` with a thread number equal to the machine cores. The fix basically uses the same mechanisms, calling `try_reserve` after the restore, both from `view` and `load` methods.

I also added a regression test to avoid this bug from happening again.

**NOTE**: the c++ might be ugly, I'm not a c++ dev. However, I'm happy to help and fix any issue.


ashvardanian:
That makes sense. I guess we can do it be default in all environments that don't support multithreading natively - also in JavaScript.

ashvardanian:
Yes, that's the one!

Stefano-t:
I've just updated the javascript part with:
- support for `threads` parameter in `exactSearch`
- support for `threads` parameter in `Index.search`
- tests for `exactSearch`, `Index.search` with different threads

Furthemore the load/view tests now use a search with multiple threads by default.

Let me know if you need anything else!

Stefano-t:
Just saw that some code that I wrote inside this branch has been added in the `main` branch, mainly in the javascript section. I updated code and removed conflicts.

Stefano-t:
OK @ashvardanian, sounds like a major improvement! Looking forward to use it!

So, we can consider this branch as stall, since the new refactoring will remove the error.

If you want, I (or you) can close this PR.

Just dropping here a quick workaround for anyone with the same issue:

run a fake search first with `threads=0` right after loading the index, the you can set the desideried thread.

ashvardanian:
Thank you, @Stefano-t! I've manually rebased your branch into `main-dev`, rewording the commits to be aligned with the rest of the commit history. Closing this one, and will release soon 🤗

Issue #629, opened 2025-06-29
Bug: Using a custom metric results in a segfault with the Rust binding


couguiv:
### Describe the bug

The `change_metric` API does not work: adding more than one vector to an Index with a custom metric results in a segfault. 

The issue is related to how the Rust closure is passed to the C++ side: `transmute`  a closure `Box<dyn Fn...>` to a function pointer `fn...` doesn't seem to work.

### Steps to reproduce

Simply add the same vector again at the end of the existing test (`test_change_distance_function`)

```
#[test]
fn test_change_distance_function() {
    let mut options = IndexOptions::default();
    options.dimensions = 2; // Adjusted for simplicity in creating test vectors
    let mut index = Index::new(&options).unwrap();
    index.reserve(10).unwrap();

    // Adding a simple vector to test the distance function changes
    let vector: [f32; 2] = [1.0, 0.0];
    index.add(1, &vector).unwrap();

    // Stateful distance function with adjustments for pointer to slice conversion
    let first_factor: f32 = 2.0;
    let second_factor: f32 = 0.7;
    let stateful_distance = Box::new(move |a: *const f32, b: *const f32| unsafe {
        let a_slice = std::slice::from_raw_parts(a, 2);
        let b_slice = std::slice::from_raw_parts(b, 2);
        (a_slice[0] - b_slice[0]).abs() * first_factor
            + (a_slice[1] - b_slice[1]).abs() * second_factor
    });
    index.change_metric(stateful_distance);

    // Add the same vector again to trigger the SEGFAULT
    index.add(2, &vector).unwrap();
}
```

### Expected behavior

It should be possible to add more than one vector to an index with a changed metric.

### USearch version

v2.17.12

### Operating System

Ubuntu 24.04.2 LTS (WSL)

### Hardware architecture

x86

### Which interface are you using?

Other bindings

### Contact Details

couguiv@protonmail.com

### Are you open to being tagged as a contributor?

- [x] I am open to being mentioned in the project `.git` history as a contributor

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

couguiv:
I actually fixed it, I will draft a PR soon.
Also, https://github.com/unum-cloud/usearch/issues/515 may be related to this bug.

ashvardanian:
Thanks, @couguiv!

ashvardanian:
[...]

PR #635, opened 2025-07-06
Fix: Properly deallocate the closure when dropping the Index

couguiv:
Store the closure's raw pointer instead of the Box.
This way the Box can be reconstructed and deallocated when dropping the wrapper Index.

This is needed since the raw pointer passed to C++ depends on the Rust closure.
So if the closure is dropped while the index is still in use, it will result in a segfault when calling the custom metric.

This behavior is a dangling pointer and/or memory leak.

Related to https://github.com/unum-cloud/usearch/issues/629


ashvardanian:
Hi @couguiv! Any addition CI tests you think we should add to check for those leaks and other issues?

couguiv:
Hello @ashvardanian !
The memory leak was initially introduced by https://github.com/unum-cloud/usearch/pull/632 to fix a bug with custom metric call.
The leak itself is negligible, it can occurs in very specific scenarios but Rust allowed me to pinpoint it.
Better (memory) safe than sorry, right ? 😄

So I can't think of any CI tests right now but maybe some Rust tests that can be improved, I will think about it.

Anyway, the initial bug is fixed. (I have been using the fixed version for 3 weeks now, it has been stable)
Thank you for your work ! 🤗

Issue #645, opened 2025-08-08
Bug: index_get::isolate and index_dense_gt::isolate are not working correctly.


yoonseok-kim:
### Describe the bug

I think `index_gt::isolate` is incorrectly implemented. There is a bug in the logic that updates incoming links from neighbors: it attempts to access elements after clearing `neighbors`

https://github.com/unum-cloud/usearch/blob/12304aec5ad9b1e9e635fdbc96b6bdbe54dc716e/include/usearch/index.hpp#L3658-L3660

-------

Also, I think `index_dense_gt::isolate` is incorrectly implemented. `index_gt::isolate` requires an `allow_member` predicate parameter, but `index_dense_gt::isolate` passes `disallow` lambda as argument of `index_gt::isolate`

https://github.com/unum-cloud/usearch/blob/12304aec5ad9b1e9e635fdbc96b6bdbe54dc716e/include/usearch/index_dense.hpp#L1712-L1717

### Steps to reproduce

I wrote a test case for index_dense_gt::isolate as below, but it failed.

```C++
// default connectivity: 16
void test_isolate() {
    constexpr std::size_t dataset_count = 16;
    constexpr std::size_t dimensions = 32;
    metric_punned_t metric(dimensions, metric_kind_t::cos_k);

    std::random_device rd;
    std::mt19937 gen(rd());
    std::uniform_real_distribution<> dis(0.0, 1.0);
    using vector_of_vectors_t = std::vector<std::vector<float>>;

    vector_of_vectors_t vector_of_vectors(dataset_count);
    for (auto& vector : vector_of_vectors) {
        vector.resize(dimensions);
        std::generate(vector.begin(), vector.end(), [&] { return dis(gen); });
    }

    index_dense_t index = index_dense_t::make(metric);
    index.reserve(dataset_count);
    for (std::size_t idx = 0; idx < dataset_count; ++idx) {
        index.add(idx, vector_of_vectors[idx].data());
    }
    expect_eq(index.size(), dataset_count);

    for (std::size_t idx = 0; idx < dataset_count; ++idx) {
        if (idx % 2 == 0)
            index.remove(idx);
    }

    auto result = index.isolate();
    for (std::size_t idx = 0; idx < dataset_count; ++idx) {
        auto result = index.search(vector_of_vectors[idx].data(), 16);
        expect_eq(result.size(), dataset_count / 2);
    }
}
```

### Expected behavior

`index_gt::isolate` is incorrectly implemented.

### USearch version

[...]

PR #646, opened 2025-08-08
Fix index_gt::isolate, index_dense_gt::isolate

yoonseok-kim:
This PR fixes https://github.com/unum-cloud/usearch/issues/645.

### DIFF
--- a/cpp/test.cpp
+++ b/cpp/test.cpp
@@ -1144,6 +1144,41 @@ void test_filtered_search() {
     }
 }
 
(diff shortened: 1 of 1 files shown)


ashvardanian:
Thanks for the PR, @yoonseok-kim! There is definitely not enough test coverage for those features, and the `index_dense_gt` fix is on-point! The `index_gt` I'd patch in a slightly different fashion. I'll polish it a bit and merge into `main` a bit later 😉

PR #653, opened 2025-08-31
javascript: improve typings

rockwotj:
Add can take a matric due to normalizeVectors being called.

Additionally, support more strict typing in `search`: if you only search with a single vector input, you get a Matches object back. This simplifies the API instead of having to assert on the return type as a user of the library.


ashvardanian (approved):

ashvardanian:
Thank you, @rockwotj!

PR #654, opened 2025-09-03
Broader GoLang API

cpegeric:
Currently, golang API only support float32 as input type.  This PR add new interfaces such as AddWithPointer, SearchWithPointer with unsafe.Pointer instead of []float32 in order to support all types.


ashvardanian:
Hi @cpegeric! Thanks for suggesting those extensions! I'll refactor them a bit and merge into `main-dev` a bit later

Issue #574, opened 2025-03-20
Feature: Follow Rust safety/soundness conventions, add `_unchecked` methods


tnibler:
### Describe what you are looking for

Hi, I just had to look up for the fourth time probably why adding to an `Index` caused a segfault. Turns out it's because you have to preallocate space using `reserve`.

This is very surprising behavior and unsound, since safe Rust should never cause undefined behavior such as out of bounds accesses.

In #412, it was mentioned that checking for sufficient capacity is not zero-cost, since it's an atomic read that could cause contention. That's fair, but also probably a rare occurrence in most usage patterns. So what do you think about basically duplicating the API into `unsafe` `*_unchecked` methods that don't perform any validation and regular safe functions that perform checks e.g., for sufficient capacity in this case? The standard library and many other lower-level libraries also use this pattern.

This library is really great, but as it is right now any project that uses it is unsound and can violate memory safety in code that is not marked `unsafe`, which is (almost) the whole point of Rust.

### Can you contribute to the implementation?

- [x] I can contribute

### Is your feature request specific to a certain interface?

Other bindings

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #670, opened 2025-10-28
Breaking revamp of Rust bindings for v3

tnibler:
Related to #574 , draft for the v3 Rust bindings. If we're gonna break APIs, let's just go all the way :D

The goal is to prevent unsoundness from the safe Rust interface, turning as many illegal operations as possible into compile errors , and catch the others at runtime.

Sorry for the mega-PR, but I don't know if it makes sense to split this up since all changes kind of affect everything.

## Error handling, type safety

In general: existing methods on `VectorType` become `unsafe fn xxx_unchecked`, and safe variants are added that check all required invariants.

There's some ambiguity as to what is actually invalid/unsafe here. As far as I can tell, using an Index created with one `ScalarKind` (say `int8`) as one of a different type (like inserting an `f64` vector) does not actually cause any memory unsafety. It can garble data though, which I'd say is enough to make the operation `unsafe` (pretty common in the ecosystem AFAIK?).

So I've added a check that the casts that will be done on the C++ make sense, but it would be good to have more opinions on which ones are ok and which ones are almost certainly mistakes. Some are easy (`f32` to `f16`, `bf16` is ok, `f64` to `b1x8` is not for example), others less so.

## Safe `Index` views into file, buffer

Currently you can have an `Index` reading a buffer passed in by the user, which will hopefully stay live as long as the `Index`. Though the method is marked unsafe, it's easy to create a dangling pointer situation which is unfortunate in Rust of all places. Also, these immutable `Index` views will throw exceptions if any mutating methods are called on them.

My proposed solution is to separate the regular, mutable `Index` that owns its memory and the immutable `IndexView` into different types. This way `IndexView` can be associated with a lifetime so the backing buffer can't be dropped while it's being used, and the `IndexView` object (holding a pointer) can not be moved between threads.

While we're at it, splitting mutating and readonly methods on `Index` into different traits allows `IndexView` to only expose read methods and prevent some runtime exceptions this way.

## Custom metric closure shenanigans

This part could maybe be split into a different PR. First a lot of duplicated code can be written in a generic way, which is always nice. There's also a memory leak pointed out in #629 (fixed by properly dropping the currently set metric closure when it's overwritten).

I only noticed that because I wanted to see how ergonomics could be improved here. Currently custom metrics are boxed closures taking `*const T` arguments, which would be very nice to instead turn into `&[T]` slices of correct length/dimensionality. Also, closures are currently double-boxed because `Box<dyn Fn..>` is a wide pointer, and `metric_punned_t` wants a regular pointer, so there's three (if I'm counting correctly) pointer dereferences per call to the metric function (trampoline -> pointer to wide pointer -> dynamic dispatch for `dyn Fn` trait object). 

I *think* it's possible to add a shim in there to turn the pointers into slices without adding another indirection (currently `closure_address` is a pointer to the `Box<dyn>`, when it could instead point to the trait object/wide pointer instead and the trampoline casts that into a valid trait object reference). But I'm not sure if it's worth it when you could also just allow the user a regular `fn (&[T], &[T]) -> Distance` pointer, which avoid a lot of that and covers many (most?) use cases just fine.

FFI and closures is a difficult topic though so it would be cool for other to weigh in and check what I'm saying is correct.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/rust/lib.rs
+++ b/rust/lib.rs
@@ -2122,6 +2122,34 @@ mod tests {
     }
 
     #[test]
+    fn test_change_metric_leak() {
+        let options = IndexOptions {
+            dimensions: 2,
+            quantization: ScalarKind::F32,
+            ..Default::default()
+        };
+        let mut index = Index::new(&options).unwrap();
+        index.reserve(10).unwrap();
+
+        let vector: [f32; 2] = [1.0, 0.0];
+        index.add(1, &vector).unwrap();
+
+        let counter = std::sync::Arc::new(std::sync::Mutex::new(0f32));
+
+        let n: i32 = 100;
+        for _ in 0..n {
+            let counter_copy = counter.clone();
+            let metric =
+                Box::new(move |_: *const f32, _: *const f32| *counter_copy.lock().unwrap());
+            index.change_metric(metric).unwrap();
+        }
+        drop(index);
+        // Only one reference to counter (the one held in this scope, not the closure)
+        // should be left.
+        assert_eq!(std::sync::Arc::strong_count(&counter), 1);
+    }
+
+    #[test]
     fn test_binary_vectors_and_hamming_distance() {
         let index = Index::new(&IndexOptions {
             dimensions: 8,
--- a/rust/lib.rs
+++ b/rust/lib.rs
@@ -431,6 +431,8 @@ pub mod ffi {
     }
 }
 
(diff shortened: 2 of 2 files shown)


flake.nix
ashvardanian: I'm not familiar with Flake. Is it a good idea to include this into the project history including with `.direnv/**` files?

tnibler: You're right it is not haha :D, I'm just not used to nix enough and forgot they weren't in gitignore. The problem is that the `flake.nix` won't work unless it's at least staged in the repo, so for non-nix projects you have to be a bit careful which I was not here.

Issue #698, opened 2026-01-07
Bug: Duplicate neighbor slots produced during incremental indexing(update, remove) in HNSW reverse links


yoonseok-kim:
### Describe the bug

There is a bug in the reverse-link formation logic that can produce duplicate neighbor entries in HNSW graphs.

In form_reverse_links_, when the neighbor list size is below connectivity_max, the new slot is appended without checking whether it already exists:

https://github.com/unum-cloud/USearch/blob/7306bb446be5f0f0c529ec8acdc57361cef8a8a7/include/usearch/index.hpp#L3899-L3904

In incremental update scenarios(frequent update + remove), the same slot may be inserted multiple times. (despite the comment stating that nothing should be done if the slot already exists.)

For reference, It is likely that this issue does not occur when the number of connections of a neighbor node exceeds M (or 2M). In such cases, the refinement(heuristic) process(refine_) is expected to naturally eliminate duplicate slots.

#### Fix

Adding a uniqueness check in this branch prevents duplicate neighbor entries.

```C++
if (close_header.size() < connectivity_max) {
    if (std::all_of(close_header.begin(), close_header.end(),
                    [new_slot](compressed_slot_t slot) {
                        return slot != new_slot;
                    })) {
        close_header.push_back(new_slot);
    }
    continue;
}
```

I have submitted a bugfix PR(https://github.com/unum-cloud/USearch/pull/699) for this issue. I would appreciate it if you could review it.

### Steps to reproduce

Perform repeated update(with different vector values) and remove operations on the same slot in index_dense_gt. This would leads to a situation where the same slot can be inserted multiple times during reverse-link formation.
(In my case, I reproduced the issue by implementing deletion manually on top of index_gt, but it is expected to occur in index_dense_gt as well.)

### Expected behavior

No duplicate slots should exist in the neighbor list.

### USearch version

v2.23.0

### Operating System

Red Hat Enterprise Linux release 8.6

### Hardware architecture

x86

### Which interface are you using?

C++ implementation

### Contact Details

kys910524@gmail.com

### Are you open to being tagged as a contributor?

- [x] I am open to being mentioned in the project `.git` history as a contributor

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

yoonseok-kim:
Thank you for merging the [PR#699](https://github.com/unum-cloud/USearch/pull/699). :)

PR #699, opened 2026-01-07
Fix: Avoid duplicate neighbor slots in HNSW reverse links

yoonseok-kim:
This PR fixes https://github.com/unum-cloud/USearch/issues/698.

### DIFF
--- a/include/usearch/index.hpp
+++ b/include/usearch/index.hpp
@@ -3899,7 +3899,8 @@ class index_gt {
             // If `new_slot` is already present in the neighboring connections of `close_slot`
             // then no need to modify any connections or run the heuristics.
             if (close_header.size() < connectivity_max) {
-                if (std::find(close_header.begin(), close_header.end(), new_slot) == close_header.end()) {
+                if (std::find_if(close_header.begin(), close_header.end(),
+                                 [new_slot](compressed_slot_t slot) { return slot == new_slot; }) == close_header.end()) {
                     close_header.push_back(new_slot);
                 }
                 continue;


yoonseok-kim:
I understand this problem may not be a critical issue, but I think it could be considered a bug. I would appreciate it if you could review whether it might be possible to address this in next release. If this PR is merged, it would benefit our product. Thank you for your consideration. :)

ashvardanian:
Hi @yoonseok-kim! Shouldn't we instead use `std::find`? It typically vectorizes much better 🤗

yoonseok-kim:
Good point. thanks! I’ll apply it.

include/usearch/index.hpp
```diff
@@ -3899,7 +3899,10 @@ class index_gt {
             // If `new_slot` is already present in the neighboring connections of `close_slot`
             // then no need to modify any connections or run the heuristics.
             if (close_header.size() < connectivity_max) {
-                close_header.push_back(new_slot);
+                if (std::find_if(close_header.begin(), close_header.end(),
```
yoonseok-kim, line 3902: `close_header`'s iterator returns `misaligned_ref_gt`, not `compressed_slot_t`, so the internal `*it == value` comparison in `std::find` fails. I tried changing to `std::find_if` so that the predicate(lambda) receives a value. when the lambda is called with `compressed_slot_t` slot, the `misaligned_ref_gt` is converted to a value and `slot == new_slot` compiles.

yoonseok-kim:
I would appreciate it if you could review this.
(I think current CI failures don't appear to be related to this PR.)

PR #704, opened 2026-01-15
Fix: Match Rust move semantics

madmax983:
fix(rust): call reserve() after heap allocation to prevent segfaults

Calling reserve() on stack-allocated index before moving to heap stores dangling pointers, causing crashes when Index is used in Rust structs, Box, Arc, or any moved context.

Move reserve() call to wrap() function after heap allocation.

Just as a reference this came up with trying to implement usearch in https://github.com/madmax983/GallifreyDB to replace hnsw_rs.

### DIFF
--- a/rust/lib.cpp
+++ b/rust/lib.cpp
@@ -211,8 +211,10 @@ std::unique_ptr<NativeIndex> new_native_index(IndexOptions const& options) {
     index_dense_config_t config(options.connectivity, options.expansion_add, options.expansion_search);
     config.multi = options.multi;
     index_t index = index_t::make(metric, config);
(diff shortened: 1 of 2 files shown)


madmax983:
Reworked it, let me know if this way is better?

ashvardanian:
Thanks, @madmax983! I'll double check in a couple of days and will be happy to merge if everything is optimal 🤗

ashvardanian:
Thank you, @madmax983! Sorry it took a while, was busy refining the [NumKong](https://github.com/ashvardanian/NumKong) CI & binding layers while the core logic was fresh in memory 🤗

PR #709, opened 2026-02-04
Change warning to pragma message for MSVC comp

fran6co:


ashvardanian:
Thank you, @fran6co! I'll need to check that it doesn't break the builds for other compiler, but seems like we may be able to merge that. Do you have any details about the original issue you are trying to solve with this?

fran6co:
Hi, I was hitting this when building with vcpkg. There is nothing preventing the code reach that place using a compiler that is not MSVC (that I think it was the intended target). I think that normally doesn't happen and probably this fix is not really needed

PR #713, opened 2026-02-16
Add: `radius` parameter for distance-bounded search

andrea-dagostino:
## Summary

- The change implements the `radius` argument for search across the C++ core and Python wrapper, allowing users to exclude results beyond a maximum distance threshold.
- Adds `float radius` to `index_search_config_t`, threaded through `index_dense_gt::search()` and `filtered_search()` overloads, with post-sort filtering in `index_gt::search()`.
- Wires the existing-but-unused `radius` parameter in Python `Index.search()` through `_search_in_compiled()`, and adds it to `Indexes.search()` and the module-level `search()`.
- Results with `distance > radius` are excluded; `distance == radius` is included. Default is infinity (no filtering, fully backward-compatible).

## Changed files

| File | What |
|---|---|
| `include/usearch/index.hpp` | `radius` field in `index_search_config_t` + filtering after `sort_ascending()`/`shrink()` |
| `include/usearch/index_dense.hpp` | `radius` param on all 12 `search()`/`filtered_search()` overloads + `search_()` |
| `python/usearch/index.py` | `radius` wired through `_search_in_compiled()`, `Index.search()`, `Indexes.search()`, `search()` |
| `cpp/test.cpp` | `test_radius_search()` -- default, tight, zero, and infinite radius with L2sq |
| `python/scripts/test_index.py` | 4 test functions (11 parametrized cases) covering filtering, defaults, edge cases, single/batch |

## Design notes

- **C++ filtering** happens after HNSW traversal completes (post-filter on sorted results), not during traversal;
- **Python filtering** uses `np.searchsorted` on the ascending-sorted distance arrays in `_search_in_compiled()`. This handles both the HNSW and brute-force exact search paths without changing `lib.cpp`;
- `lib.cpp` (pybind11 bindings) is intentionally **not changed** -- the module-level `search(exact=True)` path forwards `**kwargs` to `_exact_search`, which doesn't accept `radius`. Extracting `radius` as a named parameter in `_search_in_compiled()` avoids this cleanly.

## Tests

- [x] C++ `test_radius_search()` passes (default, tight, zero, infinite radius)
- [x] C++ full test suite passes (`build_debug/test_cpp`)
- [x] Python radius tests pass (11/11 cases)
- [x] Python full test suite passes (593 passed, 6 skipped -- skips are pre-existing)

### DIFF
--- a/cpp/test.cpp
+++ b/cpp/test.cpp
@@ -1144,6 +1144,68 @@ void test_filtered_search() {
     }
 }
 
+
+void test_radius_search() {
+    constexpr std::size_t dataset_count = 256;
+    constexpr std::size_t dimensions = 32;
+    metric_punned_t metric(dimensions, metric_kind_t::l2sq_k);
+
+    std::random_device seed_source;
+    std::mt19937 generator(seed_source());
+    std::uniform_real_distribution<float> distribution(0.0, 1.0);
+    using vector_of_vectors_t = std::vector<std::vector<float>>;
+
+    vector_of_vectors_t vector_of_vectors(dataset_count);
+    for (auto& vector : vector_of_vectors) {
+        vector.resize(dimensions);
+        std::generate(vector.begin(), vector.end(), [&] { return distribution(generator); });
+    }
+
+    index_dense_t index = index_dense_t::make(metric);
+    index.reserve(dataset_count);
+    for (std::size_t idx = 0; idx < dataset_count; ++idx)
+        index.add(idx, vector_of_vectors[idx].data());
+    expect_eq(index.size(), dataset_count);
+
+    // Search without radius (default) - should return up to `wanted` results
+    auto result_default = index.search(vector_of_vectors[0].data(), 10);
+    expect(result_default);
+    expect(result_default.count > 0);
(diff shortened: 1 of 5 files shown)


andrea-dagostino:
Any updates?

shankar-iyer:
> C++ filtering happens after HNSW traversal completes (post-filter on sorted results), not during traversal;

Would it mean that callers will have to experiment or pass a very high `nn` value to be sure that they retrieved all the nieghbours _within a_ specific distance bound?

andrea-dagostino:
> > C++ filtering happens after HNSW traversal completes (post-filter on sorted results), not during traversal;
> 
> Would it mean that callers will have to experiment or pass a very high `nn` value to be sure that they retrieved all the nieghbours _within a_ specific distance bound?

Yes. At the moment, this is by design. 

 The current flow is:
- HNSW traversal runs to find wanted neighbors
- Results are sorted ascending by distance
- Then results beyond radius are filtered out

If we don't want this, we could let radius drive the HNSW traversal termination (stopping expansion when the nearest unvisited candidate exceeds radius) rather than acting as a post-filter.

This is more of a design constraint more than other.

PR #724, opened 2026-03-06
OpenMP errors from `build.rs`

anaslimem:
## What does this PR do?

Resolves a `clippy::panicking_unwrap` lint error in the Rust `build.rs` script. 

Previously, if compiling the `usearch` C++ backend failed ( due to missing system headers like `<omp.h>` on macOS), the build script would unconditionally call `.unwrap()` on the resulting `Err`. This caused a generic panic that obscured the actual underlying compiler error and triggered a `clippy` warning.

## Why is this necessary?

Unwrapping an `Err` value is [discouraged by `cargo clippy`](https://rust-lang.github.io/rust-clippy/master/index.html#/panicking_unwrap) because it leads to unhelpful error messages:
```text
error: this call to `unwrap()` will always panic
   --> build.rs:113:17
    |
...
113 |                 result.unwrap();
    |                 ^^^^^^^^^^^^^^^
```

This PR replaces the `.unwrap()` with an explicit `panic!` that surfaces the exact `cxx` ToolExecError message. This makes debugging C++ dependency issues during `cargo build` significantly easier for developers and users.

## Testing

Verified locally that:
- `cargo clippy --all-targets --all-features` passes with zero errors/warnings.
- `cargo test -p usearch` succeeds with all 14 tests passing.
- Building the crate on a system missing `libomp` now cleanly displays the `fatal error: 'omp.h' file not found` message instead of a generic unwrap tracing panic.


ashvardanian:
Thank you, @anaslimem! The new post-NumKong build mechanism has already changed and leverages a similar idea, but I realize Clippy should probably be made part of the CI and OpenMP handling in Rust should be done differently 🤗

PR #727, opened 2026-03-10
Fix: hash set iterator returns deleted tombstones in equal_range

### LINKED ISSUE #697: Bug: Integer underflow in Index::size() during concurrent add/remove.
### Describe the bug

When performing concurrent add and remove operations, the Index::size() method can return an extremely large value (e.g., `18446744073709551614`).
This occurs because the underlying C++ implementation calculates size as: `return typed_->size() - free_keys_.size();`.
Under high concurrency, it is possible for the `free_keys_` size to be larger than the `typed_` size, causing a std::size_t underflow.

Debugging reveals the following state during the crash:
- `typed_->size()`: 88
- `free_keys_.size()`: 91
- Result: 88 - 91 = -3  -> 18446744073709551613 (approx).

### Steps to reproduce

This issue was reproduced using the Rust bindings with the tokio multi-threaded runtime. Note that even when using index.read() (assuming internal thread-safety of usearch), the inconsistency between the two internal counters is exposed.
```rust
#[tokio::test(flavor = "multi_thread")]
async fn test_concurrent_usearch_add_remove_size() {
    let options = IndexOptions {
        dimensions: 3,
        ..Default::default()
    };

    let index = Arc::new(RwLock::new(
        usearch::Index::new(&options).expect("failed to create index"),
    ));

    const VECTORS: usize = 50;
    let threads = tokio::runtime::Handle::current().metrics().num_workers();

    index.write().unwrap().reserve(VECTORS * 10).expect("failed to reserve");

    let mut handles = vec![];

    for _ in 0..threads {
        let idx = Arc::clone(&index);
        // Task 1: Concurrent Adds
        handles.push(tokio::spawn(async move {
            for i in 0..VECTORS {
                let _ = idx.read().unwrap().add(i as u64, &[i as f32; 3]);
            }
        }));

        // Task 2: Concurrent Removes
        let idx = Arc::clone(&index);
        handles.push(tokio::spawn(async move {
            for i in 0..VECTORS {
                let _ = idx.read().unwrap().remove(i as u64);
            }
        }));

        // Task 3: Monitor Size
        let idx = Arc::clone(&index);
        handles.push(tokio::spawn(async move {
            for _ in 0..VECTORS * 10 {
                let size = idx.read().unwrap().size();
                assert!(size <= VECTORS * threads, "Underflow detected! Size: {}", size);
                tokio::task::yield_now().await;
            }
        }));
    }

    for handle in handles {
        handle.await.expect("Task failed");
    }
}
```

### Expected behavior

Index::size() to always return valid number of vectors inside the index.

### USearch version

2.23.0

### Operating System

Fedora release 42 (Adams)

### Hardware architecture

x86

### Which interface are you using?

Other bindings

### Contact Details

_No response_

### Are you open to being tagged as a contributor?

- [x] I am open to being mentioned in the project `.git` history as a contributor

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DavIvek:
The equal_range iterator in `flat_hash_multi_set_gt` did not check the deleted flag when advancing. After add/remove cycles that create tombstones with the same key, the iterator would yield deleted entries as live matches. In `index_dense_gt::remove()` this caused already-freed slots to be pushed to `free_keys_` again, inflating `free_keys_.size()` and producing unsigned underflow in `size()`.

Reproduces in both multi and non-multi mode (single-threaded) whenever two keys hash-collide and one tombstone is reused by a different key.

Closes #697

### DIFF
--- a/include/usearch/index_plugins.hpp
+++ b/include/usearch/index_plugins.hpp
@@ -2735,10 +2735,17 @@ class flat_hash_multi_set_gt {
 
         // Pre-increment
         equal_iterator_gt& operator++() {
+            auto should_skip = [&](std::size_t i) -> bool {
+                auto slot = parent_->slot_ref(i);
+                if (!(slot.header.populated & slot.mask))
+                    return false; // empty slot — stop
+                if (slot.header.deleted & slot.mask)
+                    return true; // tombstone — skip
+                return !equals_(slot.element, query_); // skip non-matching live entries
+            };
             do {
                 index_ = (index_ + 1) & (parent_->capacity_slots_ - 1);
-            } while (!equals_(parent_->slot_ref(index_).element, query_) &&
-                     (parent_->slot_ref(index_).header.populated & parent_->slot_ref(index_).mask));
+            } while (should_skip(index_));
             return *this;
         }
 
--- a/include/usearch/index_plugins.hpp
+++ b/include/usearch/index_plugins.hpp
@@ -2733,19 +2733,17 @@ class flat_hash_multi_set_gt {
                           equals_t const& equals)
             : index_(index), parent_(parent), query_(query), equals_(equals) {}
 
-        // Pre-increment
+        // Pre-increment: advance past tombstones and non-matching entries,
+        // stopping at the next matching live entry or an empty slot.
         equal_iterator_gt& operator++() {
-            auto should_skip = [&](std::size_t i) -> bool {
-                auto slot = parent_->slot_ref(i);
-                if (!(slot.header.populated & slot.mask))
-                    return false; // empty slot — stop
-                if (slot.header.deleted & slot.mask)
-                    return true; // tombstone — skip
-                return !equals_(slot.element, query_); // skip non-matching live entries
-            };
             do {
                 index_ = (index_ + 1) & (parent_->capacity_slots_ - 1);
-            } while (should_skip(index_));
+                auto slot = parent_->slot_ref(index_);
+                bool is_empty = ~slot.header.populated & slot.mask;
+                bool is_match = !(slot.header.deleted & slot.mask) && equals_(slot.element, query_);
+                if (is_empty || is_match)
+                    break;
+            } while (true);
             return *this;
         }
 


DavIvek:
I was able to reproduce corruption with next test:

```
#include <usearch/index_plugins.hpp>

namespace {

struct KeySlot {
    std::uint64_t key;
    std::uint64_t slot;
};

// Guaranteed collisions.
struct CollidingHash {
    using is_transparent = void;
    std::size_t operator()(KeySlot const &) const noexcept { return 0; }
    std::size_t operator()(std::uint64_t) const noexcept { return 0; }
};

struct KeySlotEqual {
    using is_transparent = void;
    bool operator()(KeySlot const &a, std::uint64_t b) const noexcept { return a.key == b; }
    bool operator()(std::uint64_t a, KeySlot const &b) const noexcept { return a == b.key; }
    bool operator()(KeySlot const &a, KeySlot const &b) const noexcept { return a.key == b.key; }
};

using HashSet = unum::usearch::flat_hash_multi_set_gt<KeySlot, CollidingHash, KeySlotEqual>;

}  // namespace

// Simulates index_dense_gt's add/remove cycle and shows size() corruption.
TEST(UsearchTombstoneBug, SizeCorruption) {
    HashSet slot_lookup;

    // Simulated counters (mirrors index_dense_gt internals)
    std::size_t typed_size = 0;
    std::vector<std::uint64_t> free_keys;  // simulated ring buffer

    auto simulated_size = [&]() -> std::int64_t {
        return static_cast<std::int64_t>(typed_size) - static_cast<std::int64_t>(free_keys.size());
    };

    auto simulate_add = [&](std::uint64_t key) {
        std::uint64_t slot;
        if (!free_keys.empty()) {
            slot = free_keys.back();
            free_keys.pop_back();
        } else {
            slot = typed_size++;
        }
        slot_lookup.try_emplace({key, slot});
    };

    auto simulate_remove = [&](std::uint64_t key) {
        auto [begin, end] = slot_lookup.equal_range(key);
        for (auto it = begin; it != end; ++it) {
            free_keys.push_back((*it).slot);
        }
        slot_lookup.erase(key);
    };

    // Build the toxic pattern: [live-K, tombstone-K] in the same cluster.
    //
    // 1. add(J)    → [pos 0: {J, slot0}]
    // 2. add(K)    → [pos 0: {J, slot0}, pos 1: {K, slot1}]  (collision)
    // 3. remove(J) → [pos 0: {J, slot0}☠, pos 1: {K, slot1}]
    // 4. remove(K) → [pos 0: {J, slot0}☠, pos 1: {K, slot1}☠]
    // 5. add(K)    → reuses tombstone at pos 0: [pos 0: {K, slot0}, pos 1: {K, slot1}☠]
    //                                            ^^^^^^^^ live       ^^^^^^^^ tombstone
    //
    // Now remove(K) calls equal_range(K). The buggy iterator sees BOTH the
    // live entry at pos 0 AND the tombstone at pos 1 as matches.
    // It pushes slot0 AND slot1 to free_keys_, but only slot0 was actually live.
    // slot1 was already in free_keys_ from step 4. → phantom duplicate → size() drift.

    constexpr std::uint64_t K = 1;
    constexpr std::uint64_t J = 2;

    simulate_add(J);                            // step 1
    simulate_add(K);                            // step 2
    simulate_remove(J);                         // step 3
    simulate_remove(K);                         // step 4
    simulate_add(K);                            // step 5

    ASSERT_EQ(simulated_size(), 1) << "One live entry before the critical remove";

    simulate_remove(K);                         // step 6: THE BUGGY REMOVE

    // Bug: free_keys has a phantom duplicate
    EXPECT_EQ(simulated_size(), 0)
        << "size() should be 0 after removing the only entry, but free_keys_ has "
        << free_keys.size() << " entries vs typed_size " << typed_size
        << " → size() = " << simulated_size() << " (phantom duplicates in free_keys)";
}
```

DavIvek:
I believe that failing checks aren't related with proposed changes.

DavIvek:
Hi @ashvardanian. I know things can get busy, so no rush on the review/merge.

In the meantime, I’m preparing a patch on our side (https://github.com/memgraph/memgraph) since we have a release coming up in a couple of days. I wanted to check if the proposed fix in the PR generally makes sense from your perspective before we proceed with that approach.

Thanks!

ashvardanian:
Hi @DavIvek! Sorry it takes some time to review, I'm fully consumed by ashvardanian/SimSIMD#220. I need to push it through the line before validating the PRs here 🤗

Issue #730, opened 2026-03-13
Bug: `index_gt::update` uses stale vector in `refine_` during reverse link formation


yoonseok-kim:
### Describe the bug

During `index_gt::update`, the success callback (which commits the new vector to storage) is invoked after connection reconstruction. This causes `refine_` in `form_reverse_links_` to read the old vector via `citerator_at(updated_slot)` when computing inter-neighbor distances(`inter_result_dist`), leading to incorrect neighbor pruning and degraded graph quality.

#### Cause

https://github.com/unum-cloud/USearch/blob/40d127f472e9073875566f0e9308c0302b89100a/include/usearch/index_dense.hpp#L2029-L2050

https://github.com/unum-cloud/USearch/blob/40d127f472e9073875566f0e9308c0302b89100a/include/usearch/index.hpp#L3003-L3004

When `index_dense_gt` calls `index_gt::update`, it passes an `on_success` callback that writes the new vector into `vectors_lookup_`:

```cpp
auto on_success = [&](member_ref_t member) {
    // ...
    std::memcpy(vectors_lookup_[member.slot], vector_data, metric_.bytes_per_vector());
};
```

Inside `index_gt::update`, the execution order is:

1. `search_for_one_`: uses `value` directly (correct)
2. For each level:
   - `search_to_update_`: uses `value` directly (correct)
   - `form_links_to_closest_` -> `refine_`: `updated_slot` excluded from candidates (correct)
   - `form_reverse_links_` -> `refine_`:  reads `citerator_at(updated_slot)` = old vector (incorrect)
3. `callback(at(updated_slot))` — vector committed to storage (too late)

In step 2, `form_reverse_links_` inserts the updated slot into `top_for_refine` with the correct distance (computed directly from `value`). However, `refine_` subsequently computes inter-neighbor distances between candidates using `citerator_at`, which accesses`vectors_lookup_[updated_slot]`(still holding the old vector at that point).

```cpp
// form_reverse_links_(): distance to close_slot uses value directly (correct)
top_for_refine.insert_reserved({context.measure(value, citerator_at(close_slot), metric), new_slot});

// refine_(): inter-neighbor distance uses citerator_at() (stale for updated_slot)
distance_t inter_result_dist = context.measure(
    citerator_at(candidate.slot),  // stale if candidate.slot == updated_slot
    citerator_at(submitted.slot),  // stale if submitted.slot == updated_slot
    metric);
```

[...]

PR #729, opened 2026-03-13
Fix: deduplicate neighbor slots in `form_reverse_links_` in `index_gt::update`

yoonseok-kim:
This PR fixes https://github.com/unum-cloud/USearch/issues/728.

----

- For your reference, if [PR#725](https://github.com/unum-cloud/USearch/pull/725) is merged, I can also add the reproduction test case written in [the issue](https://github.com/unum-cloud/USearch/issues/728) as a new test in index_gt_test.cpp.

PR #731, opened 2026-03-13
Fix: use fresh vector for updated slot in `refine_` in `index_get::update`

yoonseok-kim:
This PR fixes https://github.com/unum-cloud/USearch/issues/730.

---

- This PR depends on [PR#729](https://github.com/unum-cloud/USearch/pull/729) and is only valid once that is merged first.
- For your reference, if [PR#725](https://github.com/unum-cloud/USearch/pull/725) is merged, I can also add the reproduction test case written in [the issue](https://github.com/unum-cloud/USearch/issues/730) as a new test in index_gt_test.cpp.


ashvardanian:
Great patch, thanks @yoonseok-kim!

PR #733, opened 2026-03-18
Fix: Strip nested sub-directories from package distributions

arsenydubrovin:
The library and header files are located in subfolders after unzip, so the move commands were failing


ashvardanian:
Good catch, @arsenydubrovin! I think a better way to fix this is by changing the Zip shape. I'll update your PR and merge in a bit 🤗

PR #737, opened 2026-03-29
Update pyproject to include ruff rules directly and groups

Samoed:
Moved rules from `contributing` to `pyproject.toml` to make them auto-usable. Also added dependency groups for better dependency management.


ashvardanian:
Thank you, @Samoed! Sorry it took a while, was busy refining the [NumKong](https://github.com/ashvardanian/NumKong) CI & binding layers while the core logic was fresh in memory 🤗

Samoed:
Great! I also created PR for fix with mypy https://github.com/unum-cloud/USearch/pull/738, but it would require a more work and I'm not sure how much it is correct, because there is no tests.

PR #738, opened 2026-03-29
Increase typing coverage 

Samoed:
Make library fully typechexked. Integrate mypy typcheking for this. Not sure how much it is correct, because there is no tests. Maybe would be better to create tests firstly


setup.py
```diff
@@ -181,6 +181,7 @@ def get_bool_env_w_name(name: str, preference: bool) -> tuple:
 install_requires = [
     "numpy",
     "tqdm",
+    "ucall",
```
ashvardanian: Let's avoid this dependency

setup.py
```diff
@@ -229,6 +230,11 @@ def get_bool_env_w_name(name: str, preference: bool) -> tuple:
         "Topic :: Database :: Database Engines/Servers",
         "Topic :: Scientific/Engineering :: Artificial Intelligence",
     ],
+    extras_require={
+        "sklearn": ["scikit-learn>=1.0.0, <2.0.0",],
+        "graph": ["networkx"],
+        "plot": ["matplotlib"],
+    },
     include_dirs=include_dirs,
```
ashvardanian, line 237: Those are also hardly needed at the library scope

python/usearch/numba.py
```diff
@@ -25,12 +25,20 @@ def jit(
...
+    )
+    signature_f16args = types.float32(
+        types.CPointer(types.float16), types.CPointer(types.float16)
+    )
+    signature_f32args = types.float32(
+        types.CPointer(types.float32), types.CPointer(types.float32)
+    )
+    signature_f64args = types.float32(
+        types.CPointer(types.float64), types.CPointer(types.float64)
+    )
```
ashvardanian: Our lines are 120 lines wide - surprised that this is warping 🤔

setup.py
Samoed: It's used inside library and better to be included in optional dependencies https://github.com/unum-cloud/USearch/blob/40d127f472e9073875566f0e9308c0302b89100a/python/usearch/eval.py#L383

Samoed:
When I tried to build this on mac I got error
```
      error: invalid argument '-std=c++17' not allowed with 'C'
      error: command '/usr/bin/cc' failed with exit code 1
```
and I fixed it with 
```python
from setuptools.command.build_ext import build_ext as _build_ext

# Flags that are only valid for C++ compilation
cpp_only_flags = {"-std=c++17", "/std:c++17"}


class MixedBuildExt(_build_ext):
    """Build extension that applies C++-only flags only to C++ source files."""

    def build_extension(self, ext):
        original_compile = self.compiler.compile

        def patched_compile(sources, **kwargs):
            c_sources = [s for s in sources if not s.endswith((".cpp", ".cxx", ".cc"))]
            cpp_sources = [s for s in sources if s.endswith((".cpp", ".cxx", ".cc"))]
            objects = []
            extra_args = kwargs.get("extra_postargs", [])
            if c_sources:
                c_kwargs = dict(kwargs)
                c_kwargs["extra_postargs"] = [f for f in extra_args if f not in cpp_only_flags]
                objects += original_compile(c_sources, **c_kwargs)
            if cpp_sources:
                objects += original_compile(cpp_sources, **kwargs)
            return objects

        self.compiler.compile = patched_compile
        try:
            _build_ext.build_extension(self, ext)
        finally:
            self.compiler.compile = original_compile

setup(
    ...,
    cmdclass={"build_ext": MixedBuildExt},
)
```

Not sure if it is good (I asked to fix claude), but I thought you might be interested

setup.py
Samoed: Ucall is used a lot inside library. Is it expected to be an optional dependency?

PR #758, opened 2026-05-23
fix(cast_to_i8/u8): guard zero-magnitude inputs against NaN cast

desertfury:
The integer-quantizing casts normalize by `magnitude = ||v||`. An all-zero vector makes `magnitude == 0`, so `typed_input[i] * K / magnitude` produces NaN, and casting NaN to `int8_t` / `uint8_t` is undefined behavior (UBSan: float-cast-overflow). The path is reachable from any `usearch_add` against an `i8`/`u8`-quantized index with a zero input vector — the fuzzer triggered it directly.

Emits a zero output vector when magnitude is not strictly positive (also covers NaN inputs that would have made magnitude NaN).


ashvardanian:
Might be a good idea to more aggressively fuzz the NumKong kernels against similar issues, @desertfury 🤔

PR #762, opened 2026-05-23
fix(equal_iterator_gt): guard against infinite loop on a full table

desertfury:
\`equal_iterator_gt::operator++\` walked the linear-probe chain until it hit an empty or matching slot. After many add/remove cycles every slot in the table can be populated (some with the \`deleted\` flag set, which is still \`populated == 1\`), and the loop walked forever — any \`std::distance(begin, end)\` over the equal range then hung.

Tracks the start index and pins to the \`capacity_slots_\` sentinel after a full wrap so callers comparing against the end iterator terminate.


ashvardanian:
Hi @desertfury, that one was already resolved, about to merge 🤗

PR #766, opened 2026-05-26
Bundle NumKong runtime with C# native artifacts

sergey-v9:
### Summary

This updates the C# native build and packaging flow for USearch v2.25, where NumKong is now a separate runtime library instead of being embedded into `libusearch_c`.

Changes include:
- Build Windows C# native artifacts with `USEARCH_USE_NUMKONG=1`.
- Copy `libnumkong.so`, `libnumkong.dylib`, and `numkong.dll` alongside `libusearch_c`.
- Add loader/RPATH handling so `libusearch_c` can find NumKong at runtime.
- Add native runtime asset tests to catch missing bundled dependencies.

### Validation

- Built `usearch_c` locally with NumKong enabled on Windows.
- Confirmed `libusearch_c.dll` depends on `numkong.dll` via `dumpbin`.


ashvardanian:
This looks very clean, thank you @sergey-v9!

sergey-v9:
@ashvardanian Would you mind to release a new version with that changes, so we can consume the package fom a NuGet feed?

Issue #753, opened 2026-05-15
Infinite loop in flat_hash_multi_set_gt::operator++ when all slots are populated+tombstoned


bingh0:
### Describe the bug

`flat_hash_multi_set_gt::equal_iterator_gt::operator++()` in `index_plugins.hpp` (line ~3920) can enter an infinite loop when every slot in the hash table is either live or tombstoned (no never-populated slots remain).

The `do { ... } while (true)` loop only breaks on an empty slot (`~populated & mask`) or a matching live entry. When all slots are populated — which happens after many add/remove cycles — neither condition is ever true after the last match is consumed, and the loop spins forever.

```cpp
// index_plugins.hpp ~L3920
equal_iterator_gt& operator++() {
    do {
        index_ = (index_ + 1) & (parent_->capacity_slots_ - 1);
        auto slot = parent_->slot_ref(index_);
        bool is_empty = ~slot.header.populated & slot.mask;       // never true
        bool is_match = !(slot.header.deleted & slot.mask) && equals_(slot.element, query_); // false after last match
        if (is_empty || is_match)
            break;                                                // never reached
    } while (true);  // infinite loop
    return *this;
}
```

This is called from `index_dense_gt::remove()` via `std::distance(matching_slots.first, matching_slots.second)` and the subsequent `for` loop over the `equal_range` iterators.

### Steps to reproduce

1. Create an index
2. Add N entries (enough to fill the hash table's initial capacity)
3. Remove most of them (creates tombstones: `populated=1, deleted=1`)
4. Add new entries that reuse tombstoned slots
5. Repeat add/remove cycles until zero never-populated slots remain
6. Call `remove()` on any key — `operator++` spins forever

The trigger condition is: `populated_count + tombstone_count == capacity_slots_` (zero virgin slots). This happens faster with small indices and high churn.

### Workload that triggers it

AI agent memory systems with ephemeral/journaling stores — continuous insert → evict cycles on a bounded-size index. The index stays small (~hundreds to low thousands of entries) but the cumulative add/remove count far exceeds the hash table capacity, filling every slot with either a live entry or a tombstone.

We hit this in production: a Node.js process pegged a CPU core at 100% for 2+ hours, stuck in a single `index_dense_gt::remove()` call. `perf record` showed 99.96% of samples in `usearch::index_dense_gt::remove()`.

### Root cause

[...]

PR #769, opened 2026-06-01
Fix: Rehash slot_lookup_ tombstones to stop ghost entries under churn

desertfury:
### Problem
  b779c47 (#753) bounded the saturated probe but left tombstones unreclaimed, turning the CPU-spin hang into silent ghost keys.

  ### Fix
  Track tombstones and rehash to reclaim them before saturation, so live keys stay findable under churn.

  ### Tests
  `cpp/ghost_repro.cpp` now asserts zero ghosts; previously it leaked 1616 at a tight reserve.


ashvardanian:
Thank you, @desertfury! Sorry it took a bit of time. I'm finalising new NumKong/StringZilla/ForkUnion releases to feed into USearch v3 🤗

Issue #735, opened 2026-03-24
Bug: Concurrent add() creates nodes unreachable by search()


DavIvek:
### Describe the bug

When multiple threads call `add()` in parallel, some nodes become unreachable via `search()` despite being correctly stored in the index. `contains()` and `get()` confirm the nodes exist with correct data, but graph traversal never visits them.

### Steps to reproduce

```
#include <iostream>
#include <random>
#include <thread>
#include <vector>

#include <usearch/index_dense.hpp>

using namespace unum::usearch;


int main() {
    constexpr auto num_sessions = 10;
    constexpr auto vectors_per_session = 100;
    constexpr auto total_vectors = num_sessions * vectors_per_session;
    constexpr auto dimensions = 128;

    auto index_result = index_dense_gt<>::make(metric_punned_t(dimensions, metric_kind_t::cos_k, scalar_kind_t::f32_k));
    auto& index = index_result.index;
    index.try_reserve({total_vectors, num_sessions});

    // Generate vectors
    auto all_vectors = std::vector<std::vector<float>>(total_vectors);
    auto rng = std::mt19937(42);
    auto dist = std::uniform_real_distribution<float>(-1.0f, 1.0f);
    for (auto& v : all_vectors) {
        v.resize(dimensions);
        for (auto& x : v) x = dist(rng);
    }

    // Parallel insert — jthreads auto-join at scope exit
    {
        auto threads = std::vector<std::jthread>();
        for (auto s = 0; s < num_sessions; s++) {
            threads.emplace_back([&, s] {
                auto start = s * vectors_per_session;
                for (auto i = 0; i < vectors_per_session; i++)
                    index.add(start + i, all_vectors[start + i].data(), s);
            });
        }
    }

    // Verify: search for every vector requesting ALL results
    auto misses = 0;
    for (auto i = 0; i < total_vectors; i++) {
        auto sr = index.search(all_vectors[i].data(), total_vectors);
        auto keys = std::vector<std::uint64_t>(total_vectors);
        auto distances = std::vector<float>(total_vectors);
        auto found = sr.dump_to(keys.data(), distances.data());

        auto found_self = false;
        for (auto r = 0; r < found; r++)
            if (keys[r] == i) { found_self = true; break; }
        if (!found_self) misses++;
    }

    // Confirm all nodes are stored correctly
    auto contains_ok = 0;
    for (auto i = 0; i < total_vectors; i++)
        if (index.contains(i)) contains_ok++;

[...]

PR #772, opened 2026-06-25
Fix: concurrent add() no longer strands unreachable nodes (#735)

cpegeric:
Concurrent add() could leave nodes that are stored (contains()==true) yet unreachable via search(): contained-but-orphaned. Root cause is the per-node lock being released between levels while a node builds its links top-down.

A node S that had built its level-1 links but not yet its level-0 links was already discoverable (via its level-1 reverse links) while its level-0 neighbor list was still empty. A concurrent inserter greedily descending onto S as its level-0 seed dead-ended the level search with a single candidate, attached by one fragile edge, and was then evicted by the neighbor heuristic into a permanently unreachable node. The old `neighbors_(new_node,level).clear()` also actively discarded reverse links that concurrent inserters had appended to the node during the between-levels gap.

Fix, in index_gt::add_():
  - Split the single per-level loop into two passes: build ALL of the node's forward links (every level) BEFORE adding ANY reverse links. Reverse links are what make a node discoverable, so deferring them guarantees the node is never reachable as a descent seed while one of its lower levels is empty.
  - form_links_to_closest_ no longer asserts/requires a blank list: it preserves any concurrently-added reverse links and appends the forward links, deduped and bounded by the level capacity, instead of clear()+push_back. Removing the destructive clear() also stops dropping valid concurrent edges.

Single-threaded behavior is unchanged (the list is empty, so the merge path is a no-op). Adds <vector> for the reverse-link snapshot buffer.

Adds test_concurrent_add_reachability (cpp/test.cpp): builds an index from many threads and asserts every node is reachable. The pre-existing concurrent test only ran single-threaded, so this gap was uncovered.

  Validation

  - Issue repro + stress (8,000 vectors × 16 threads × 8 iterations): orphans drop from a handful per run to zero.
  - Full cpp/test.cpp suite passes, no regressions.
  - Index quality unchanged. Same data, single-threaded vs concurrent build: equal size (+0.01%), equal average degree,
  equal recall.
  - 1M scale (wiki_all, 768-dim, M=16): a concurrent build is now statistically indistinguishable from a serial one —
  true-orphan (zero-in-edge) count of 1,098/1M concurrent vs 1,123/1M single-threaded, identical recall (~91%). That
  residual ~0.11% is present identically in a serial build — it's the neighbor-selection heuristic occasionally evicting
  a node's last in-edge at scale, not this race. Concurrent builds were ~5× faster at equal quality.

  Notes

  This does not touch the update_ path (which intentionally clears and rebuilds), and it does not address the small
  algorithmic zero-in-edge fraction at low connectivity / large N, which is independent of threading.

### DIFF
--- a/cpp/test.cpp
+++ b/cpp/test.cpp
@@ -710,6 +710,74 @@ void test_punned_concurrent_updates(index_at& index, typename index_at::vector_k
 }
 
 /**
+ * Regression test for USearch #735: concurrent `add()` must never leave a node
+ * stored-but-unreachable. The bug: while a node was building links top-down, its
+ * per-level lock was released between levels, so a concurrent inserter could
+ * descend onto it before its lower level had any links, dead-end the level search,
+ * and end up attached by a single fragile edge that was then evicted — leaving a
+ * node for which `contains()` is true but `search()` can never reach it.
+ *
+ * This builds an index from many threads, then asserts every inserted vector finds
+ * itself in an exhaustive search. A single orphaned node fails the test. Before the
+ * fix this reproduced a handful of unreachable nodes per run at a few thousand
+ * vectors; with serialized (single-thread) builds it always passed — so the test is
+ * only meaningful with `threads_count > 1`.
+ *
+ * @param dimensions Number of dimensions per vector.
+ * @param vectors_per_thread Vectors inserted by each worker thread.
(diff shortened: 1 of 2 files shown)


cpegeric:
see issue #735

PR #773, opened 2026-07-03
Fix: replace identity hash with SplitMix64 for sequential integer keys

### LINKED ISSUE #770: index_dense key-lookup table degrades catastrophically on sequential/sparse integer keys (identity hash + power-of-2 linear probing) — hits both add() and view()/load (reindex_keys_)
## Summary

The key→slot lookup table inside `index_dense_gt` uses what amounts to an identity hash
for integer keys (`std::hash<u64>` is the identity function in libstdc++/libc++) over a
power-of-2-sized open-addressing table with linear probing. With keys that are ascending
(or sparse-but-ordered) integers — an extremely common pattern when keys are database row
ids or auto-increment entity ids — the table degenerates into very long contiguous
probe clusters. We hit this in production twice, in two different code paths:

1. **`add()` (build time):** key-table maintenance across incremental growth/rehash makes
   insertion throughput collapse as the index grows.
2. **`view()` / load (read time):** the table is unconditionally rebuilt via
   `reindex_keys_()` when key lookups are enabled, single-threaded — so even a
   *pre-built* index pays the full cost again on every cold start.

`enable_key_lookups = false` avoids both, but it is only reachable from some bindings
(Python exposes it; the Rust cxx bridge does not — it never sets the field, so the C++
default `true` always applies).

## Environment

- usearch **2.25.3** (Python wheel for the build; Rust crate `usearch` via the cxx bridge
  for the serving process)
- Linux x86_64, 64 GB+ RAM, index on local SSD
- Index: ~**112.3M** vectors, 384-dim, `i8`, cosine, default connectivity/expansion
  (c=16, ef_add=32, ef_search=64)
- Keys: 64-bit integers, **ascending**, sparse, max key ≈ **1e9** (so ~10% density over
  the key range, perfectly ordered insertion)

## Measured impact

| scenario | config | throughput / wall |
|---|---|---|
| `add()` build, ascending keys | default (`enable_key_lookups=true`) | degrades to ~**169 rows/s** as the index grows (build effectively never finishes — projected weeks) |
| `add()` build, same data | `enable_key_lookups=false` | ~**33,000 rows/s** sustained start-of-build, ~26,700 rows/s average over the full 112.3M; **complete in ~70 min** (16 threads) |
| controlled A/B, same slice of keys/vectors | lookups on vs off | ≈ **90×** insertion-throughput difference (~144 rows/s vs ~12,900 rows/s on the post-rehash key range) |
| `view()` of the finished 112.3M index (Rust bridge) | default (flag not settable) | **>1 hour** pegging one core inside `index_dense_gt::reindex_keys_()` before we killed it |
| `view()` of the same index | `enable_key_lookups=false` (patched bridge, see below) | ~**4 seconds** to a query-ready view |

## Where it happens (2.25.3 sources)

- `include/usearch/index_dense.hpp:138` — `index_dense_config_t::enable_key_lookups`
  defaults to `true`.
- `index_dense.hpp` (~line 1054 in 2.25.3) — `view()` path: when key lookups are enabled,
  the key table is unconditionally rebuilt via `reindex_keys_()`. Single-threaded O(N)
  rebuild, with each insert paying the clustered-probe cost — for sequential keys this is
  effectively quadratic-ish in practice.
- `rust/lib.rs` / `rust/lib.cpp` — the cxx bridge's `IndexOptions` has no
  `enable_key_lookups` member and `new_native_index` never sets
  `config.enable_key_lookups`, so Rust users always get the default `true` and cannot opt
  out, even for pure `view()+search()` consumers that never call
  `get`/`remove`/`contains`/`rename`.

## Minimal repro recipe (Python, build side)

```python
import numpy as np
from usearch.index import Index

N = 20_000_000                       # large enough to show the cliff clearly
keys = np.arange(N, dtype=np.int64) * 9 + 7   # ascending, sparse, max ~1.8e8
vecs = (np.random.rand(N, 384) * 254 - 127).astype(np.int8)

# A: default — watch rows/s collapse as the table grows/rehashes
ix = Index(ndim=384, metric="cos", dtype="i8")
ix.add(keys, vecs, log=True)

# B: identical, but
ix = Index(ndim=384, metric="cos", dtype="i8", enable_key_lookups=False)
ix.add(keys, vecs, log=True)         # orders of magnitude faster, flat
```

Read side: save the A index, then `Index.restore(..., view=True)` (or the Rust
`view()`) and time "open → first search". With lookups enabled the entire key table is
rebuilt before the view is usable.

Random (shuffled) keys do not exhibit the cliff — which is exactly the signature of
identity-hash + linear-probe clustering, not of HNSW itself.

## Suggested fixes (any one of these would resolve it; they compose)

1. **Hash the key through a cheap bit-mixer finalizer** before masking into the
   power-of-2 table (e.g. `splitmix64` finalizer or Fibonacci multiply-shift). This fixes
   sequential/sparse integer keys generically, at ~1-2 ns per lookup, with no API change.
2. **Expose `enable_key_lookups` through all bindings** — specifically the Rust cxx
   bridge (`IndexOptions`), which currently hard-inherits the C++ default. We have a
   working 3-line passthrough patch (vendored locally) and are happy to PR it:
   `IndexOptions.enable_key_lookups: bool` defaulting to `true` (behavior-preserving for
   existing callers), wired through `new_native_index`.
3. **Make `view()`/load skip or defer `reindex_keys_()`** — e.g. an explicit
   `view(reindex_keys=false)` (or lazily build the table on first `get`/`remove`). A
   serving process that only ever calls `search()` should not pay an O(N) single-thread
   rebuild on every cold start of a read-only mmap view.
4. (Minor) When the final count is known at load/build time, **reserve the key table to
   final capacity once** to avoid intermediate growth/rehash convoys.

## Workaround for others hitting this

- Python builds: pass `enable_key_lookups=False` if you never use
  `get`/`remove`/`contains`/`rename` on the index.
- Rust: not currently reachable; we vendored 2.25.3 with the passthrough described in
  (2). Note the documented trade-off: with lookups disabled, `get`/`remove`/`contains`/
  `rename` assert (`index_dense.hpp:794/808/951/1596`) — fine for add/search/view-only
  workloads.

Thanks for usearch — outside this pathology it has been excellent at this scale
(112M × 384-dim i8 in a single index, ~60 GB file, mmap-served).

chakshu-dhannawat:
Fixes #770.

## What

`std::hash<uint64_t>` is the identity function on libstdc++ and libc++. With the power-of-2 open-addressing table and linear probing in `flat_hash_multi_set_gt`, ascending integer keys (database row ids, auto-increment entity ids) produce contiguous probe clusters. The issue reporter measured a 90x throughput collapse on 112M-vector indexes.

**`include/usearch/index.hpp`** - Add `hash_gt` specializations for `std::uint64_t` and `std::int64_t` that apply the SplitMix64 bit-mixer finalizer. Overhead is about 2 ns per lookup. No API or on-disk format change.

**`rust/lib.rs` + `rust/lib.cpp`** - Expose `enable_key_lookups` through the Rust cxx bridge `IndexOptions`. Previously the field was unreachable from Rust, so all Rust callers (including `view()`/`load()`) always paid the full `reindex_keys_()` rebuild cost even for add/search-only workloads.

## Why SplitMix64

It is the standard bit-mixer for this exact problem (sequential integer keys into power-of-2 hash tables), used in Abseil's hash map and referenced by the issue author. Three multiplications and three XOR-shifts, constant time, no state.

## Backward compatibility

- Existing Python/C++ callers: no change needed. The hash function is an internal detail.
- Rust callers: `enable_key_lookups` defaults to `true`, matching the previous behavior. Callers that never use `get`/`remove`/`contains`/`rename` can now set it to `false` to skip the per-startup `reindex_keys_()` rebuild.

### DIFF
--- a/include/usearch/index.hpp
+++ b/include/usearch/index.hpp
@@ -1277,6 +1277,31 @@ template <typename element_at> struct hash_gt {
     std::size_t operator()(element_at const& element) const noexcept { return std::hash<element_at>{}(element); }
 };
 
+/**
+ *  @brief  SplitMix64 finalizer, used to scatter integer keys before masking.
+ *
+ *  On libstdc++ and libc++ `std::hash` is the identity for integers. Our open-addressing
+ *  tables mask that hash into a power-of-2 slot count and probe linearly until an @b empty
+ *  slot, so consecutive keys land in adjacent slots and merge into one contiguous run.
+ *  Both insertions and lookups then scan that run end-to-end, which is quadratic overall:
+ *  a dense ascending key range collapses insertion throughput by three orders of magnitude.
+ *  Mixing costs ~20ns per key and is dwarfed by the graph traversal in `add`.
+ */
+template <> struct hash_gt<std::uint64_t> {
+    std::size_t operator()(std::uint64_t const& element) const noexcept {
+        std::uint64_t x = element;
+        x = (x ^ (x >> 30u)) * 0xBF58476D1CE4E5B9ULL;
+        x = (x ^ (x >> 27u)) * 0x94D049BB133111EBULL;
+        return static_cast<std::size_t>(x ^ (x >> 31u));
+    }
+};
+
+template <> struct hash_gt<std::int64_t> {
+    std::size_t operator()(std::int64_t const& element) const noexcept {
+        return hash_gt<std::uint64_t>{}(static_cast<std::uint64_t>(element));
+    }
+};
+
 template <> struct hash_gt<uint40_t> {
     std::size_t operator()(uint40_t const& element) const noexcept { return std::hash<std::size_t>{}(element); }
 };


ashvardanian:
Thank you, @chakshu-dhannawat! Sorry it took a bit of time. I'm finalising new NumKong/StringZilla/ForkUnion releases to feed into USearch v3 🤗

PR #776, opened 2026-07-18
Bound file-controlled sizes in view(): levels region and dense vectors span (OOB read on crafted index)

professor-moody:
### Summary

`index_gt::view` (and `index_dense::view`) read file-controlled sizes and dereference the memory mapping before bounding them, so a crafted `.usearch` index over-reads the mapping. Two sinks:

1. **Base `view()` levels region** (`index.hpp`): `view` sizes a memory-mapped `levels[]` array from the file's `header.size` and walks it in the offset-precompute loop *before* the `if (file.size() < total_bytes)` check runs. A crafted header declaring more nodes than the file contains reads `levels[]` past the mapping. Fix: reject when the file does not extend to `first_offset` (the end of the levels array) before the loop.

2. **Dense `view()` vectors span** (`index_dense.hpp`): `vectors_buffer = {file.data() + offset, matrix_rows * matrix_cols}` forms a span with no file-size check (and the product can overflow `size_t`), and the following `if (file.size() - offset < sizeof(buffer))` guard is an unsigned subtraction that underflows once `offset` passes `file.size()`. Fix: bound the span with a checked multiply and make the header guard underflow-safe.

### Impact

Memory-mapping a crafted or untrusted `.usearch` index via `Index.view` / `Index.load(..., view=True)` yields an out-of-bounds read (crash / adjacent-memory disclosure). CWE-125.

### Validation

AddressSanitizer, control vs crafted, driving the real `view()` over a buffer-backed mapping: a well-formed index loads cleanly; a crafted one (`header.size` larger than the file / oversized matrix dims) is now rejected with a clear error and no OOB. Found with a structure-aware model/index parser fuzzer.

### DIFF
--- a/cpp/test.cpp
+++ b/cpp/test.cpp
@@ -1427,6 +1427,67 @@ void test_load_after_metric_make() {
     std::remove(path);
 }
 
+/**
+ *  @brief  Feeds `view` a truncated index file.
+ *
+ *  The matrix dimensions and the node count come from the file itself, so a short file
+ *  describes a matrix and a level array that reach past the mapping. Both must be rejected
+ *  rather than walked, and a caller-supplied offset past the end must fail the same way.
+ */
+void test_view_of_truncated_file() {
+    std::printf("Testing view of a truncated index file\n");
+
+    using index_t = index_dense_gt<std::int64_t, std::uint32_t>;
+    std::size_t const dimensions = 64;
+    std::size_t const collection = 2048;
+    char const* path = "tmp_truncated.usearch";
+
+    metric_punned_t metric(dimensions, metric_kind_t::l2sq_k, scalar_kind<f32_t>());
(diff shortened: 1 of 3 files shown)


professor-moody:
Small correction to the Impact section of this PR: I named a Python entry point that does not exist.

I wrote `Index.load(..., view=True)`. `Index.load` takes `(self, path_or_buffer, progress)` and has no `view` parameter and no `**kwargs`, so that call raises `TypeError`. Anyone reproducing from my text would hit that and reasonably wonder whether I had exercised it at all.

The entry points that actually memory-map, and therefore reach `view()`, are:

```
Index.view(path_or_buffer)
Index.restore(path_or_buffer, view=True)
```

`Index.view` alone is sufficient and is the one I should have written.

Nothing else changes. The two sinks and the fix are unaffected.

PR #780, opened 2026-07-24
Fix: treat NumPy uint8 vectors as u8, not bit-packed binary

### LINKED ISSUE #595: Bug: why are vector values altered when building index?
### Describe the bug

Apologies for a possibly silly question, I'm new to vector search.

Why does the index and and distance calculations become all zeroes?

### Steps to reproduce

```
index = Index(ndim=3)
a = np.uint8([
    [0, 0, 1],
    [0, 1, 2],
    [1, 2, 3],
])
index.add([0,1,2], a)
for i in range(3):
    print(index[i])
pd.DataFrame([r for r in index.search(a, 4)])
```

Image

### Expected behavior

If you do this with DuckDB:
```python
df = pd.DataFrame({"idx": [0,1,2], "vec": [v for v in a]})
duckdb.sql("""
SELECT a.idx, b.idx, LIST_DISTANCE(a.vec, b.vec)
FROM df a JOIN df b ON 1=1
""").df()
```

Image

### USearch version

2.17.7

### Operating System

Amazon Linux

### Hardware architecture

x86

### Which interface are you using?

Python bindings

### Contact Details

_No response_

### Are you open to being tagged as a contributor?

- [ ] I am open to being mentioned in the project `.git` history as a contributor

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

chakshu-dhannawat:
Closes #595.

### The bug

A `u8` index silently corrupts its vectors. Adding `np.uint8` data zeros out
almost everything, so search results are meaningless. `int8` works fine.

```python
import numpy as np
from usearch.index import Index, ScalarKind

vec = np.zeros(8, dtype=np.uint8); vec[2] = 3; vec[5] = 7
idx = Index(ndim=8, metric="l2sq", dtype=ScalarKind.U8)
idx.add(0, vec)
print(idx.get(0, ScalarKind.U8))   # [0 0 0 0 0 0 0 0]  <- input was [0 0 3 0 0 7 0 0]
```

### Root cause

A NumPy `uint8` buffer is ambiguous: the same bytes back both bit-packed binary
vectors (`b1x8`) and unsigned byte vectors (`u8`). `numpy_string_to_kind` maps
`uint8` to `b1x8` for backwards compatibility. When no explicit `dtype` is passed
to `add`/`search`, the dispatch used that detected kind directly, so a `u8` index
reinterpreted its data as packed bits and the values collapsed to (near-)zero.
`int8` avoids this because it maps to `i8` unambiguously.

The index already knows it is a `u8` index, so that is enough to disambiguate.

### The fix

When the caller gave no `dtype` and the buffer resolves to `b1x8` but the index's
own scalar kind is `u8`, use `u8`. Binary indexes are untouched, so bit-packed
`uint8` vectors keep working exactly as before. This is wired through a small
shared helper at the add, search, and cluster dispatch sites. The multi-shard
wrapper gains a `scalar_kind()` accessor (mirroring its existing `scalar_words()`)
so the helper compiles for both index types.

### Tests

Added `test_u8_vectors_not_misread_as_binary`: it stores sparse `uint8` counts,
checks the round trip and search ranking, and confirms a binary index is
unaffected. It fails on the current code (vector stored as all zeros) and passes
with the fix. Full `test_index.py` suite is green (609 passed, 6 pre-existing skips).

### DIFF
--- a/python/lib.cpp
+++ b/python/lib.cpp
@@ -97,6 +97,9 @@ struct dense_indexes_py_t {
     void merge(std::shared_ptr<dense_index_py_t> shard) { shards_.push_back(shard); }
     std::size_t bytes_per_vector() const noexcept { return shards_.empty() ? 0 : shards_[0]->bytes_per_vector(); }
     std::size_t scalar_words() const noexcept { return shards_.empty() ? 0 : shards_[0]->scalar_words(); }
+    scalar_kind_t scalar_kind() const noexcept {
+        return shards_.empty() ? scalar_kind_t::unknown_k : shards_[0]->scalar_kind();
+    }
     index_limits_t limits() const noexcept { return {size(), std::numeric_limits<std::size_t>::max()}; }
 
     void merge_paths(std::vector<std::string> const& paths, bool view = true, std::size_t threads = 0) {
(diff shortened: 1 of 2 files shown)


chakshu-dhannawat:
Heads up on the one red check (Ubuntu GCC 14): it's the C++ core test failing at `cpp/test.cpp:433` in `test_minimal_three_vectors` (the `std::abs(matched_distances[0]) < 0.01` self-match assertion after the serialize/reload roundtrip). That's flaky on the GCC 14 toolchain and unrelated to this PR:

- This change only touches `python/lib.cpp` and `python/scripts/test_index.py`, neither of which is compiled into the `test_cpp` binary.
- The same assertion at line 433 fails and passes nondeterministically on `main-dev` itself, e.g. run 29111716197 failed at line 433 while 29101749568 (a push about a minute earlier) passed.

The 22 other checks are green, including the Python suite that covers the actual fix. A rerun of just the GCC 14 job should clear it, but I don't have permissions to trigger one.

chakshu-dhannawat:
Hi, gentle ping for a review. Also, the GCC14 job looks flaky — would you mind re-running it? Thanks.

